PR #5, opened 2019-04-04
added git submodule initialization

### LINKED ISSUE #3: Make should pull the latest changes of submodules in the implementation folder
In order to test the latest changes that are made to Polkadot RE implementations, everytime

    make

is called, it should run
 
    git submodule update --recursive --remote

So it makes and test the latest implementations.

fgimenez:
Closes #3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,8 @@ project (PolkadotRuntimeEnvironmentTests)
 # The test helpers in various languages
 add_subdirectory(test/testers)
 
+add_custom_target(git_submodule_init ALL COMMAND git submodule update --init --recursive)
+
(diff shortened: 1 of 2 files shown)


drskalman (changes requested):
shouldn't the custom command be defined in the root CMakeLists.txt? Because it is for all implementation it is not just rust specific.

Similarly maybe it should depend on a general "build" target or something like that, so if we change the name of the this one specific target it shouldn't fail.

fgimenez:
@drskalman absolutely, done now in the root CMakeLists.txt and added as a dependency to `ALL`, PTAL.

drskalman (approved):

PR #6, opened 2019-04-04
Align Go and Rust build and install processes

### LINKED ISSUE #2: Make install should call cargo install for rust
The scripts for [CMake rust support](https://github.com/w3f/polkadot-re-tests/blob/master/cmake/CMakeCargo.cmake) runs "cargo build" fine  (when you write make it runs cargo build) but it doesn't support cargo install, when you issue "make install". So I had to write these unreliable lines:

https://github.com/w3f/polkadot-re-tests/blob/0137463f3224a8150b29184e56eb00308987f922/test/testers/rust-tester/CMakeLists.txt#L6

as temprory measure because I wanted all executables to go somewhere deterministic independent of the os, build, etc so I can run them by the Julia script. But the correct way of doing it is to modify the CMakeRust scripts so make install runs cargo install properly.

The original CMakeRust script is coming from this repo:

https://github.com/wayk/CMakeRust
:END:
nil:END:

fgimenez:
Closes #2


fgimenez:
@drskalman ready for review, please let me know WDYT.

Following with the comments in #2, now the `CMakeList.txt` file for each of the rust testers should have only one call to `cargo_build(NAME <tester>)`, in `cargo_build` there's a call to `install()` with the proper values, the same values used during the build process. This call adds rules to the `install` target, which are used when calling `make install`.

I've also refactored a bit `cmake/CMakeCargo.cmake`, extracting the set up of the envirnment variables that depend on the platform to a separate function.

WRT the go tester, currently the build command is configured to write the output to the same path as the final destination in rust, so AIUI there's no need to add `install` rules.

Finally, I've added a call to `make install` to `build_test.sh`, so that we don't need to create a separate `install_test.sh` for it.

Any thoughts are more than welcome :) cheers!

cmake/CMakeCargo.cmake
```diff
@@ -34,4 +76,6 @@ function(cargo_build)
   add_library(${CARGO_NAME} STATIC IMPORTED GLOBAL)
   add_dependencies(${CARGO_NAME} ${CARGO_NAME}_target)
   set_target_properties(${CARGO_NAME} PROPERTIES IMPORTED_LOCATION ${LIB_FILE})
+
+  install(FILES ${LIB_COMPILED_FILE} DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
drskalman: rather

    install(FILES ${LIB_COMPILED_FILE} DESTINATION bin)

is more standard

cmake/GolangSimple.cmake
```diff
@@ -10,15 +10,14 @@ function(ADD_GO_INSTALLABLE_PROGRAM NAME MAIN_SRC)
...
                     ${CMAKE_GO_FLAGS} ${MAIN_SRC}
                     WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
                     DEPENDS ${MAIN_SRC_ABS})
   foreach(DEP ${ARGN})
     add_dependencies(${NAME} ${DEP})
   endforeach()
-  
+
   add_custom_target(${NAME}_all ALL DEPENDS ${NAME})
-  install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/${NAME} DESTINATION bin)
```
drskalman: I would leave this line and then run 

     make DESTDIR=bin install  

then I'll adjust 
https://github.com/w3f/polkadot-re-tests/blob/0137463f3224a8150b29184e56eb00308987f922/test/test_fixtures.jl#L20
To point to the new correct location

    ../build/bin/usr/local/bin/scale_codec,
    ../build/bin/usr/local/bin/scale_codec_go

That's the standard behavoir of CMake with c++ which I'm gonig to add soon.

drskalman (changes requested):
Thanks a lot, few small changes and we should be good to merge.

fgimenez:
@drskalman good points, done, I've taken the opportunity to add `DESTDIR=bin` to the `make install` call in `build_test.sh`, let me know WDYT.

drskalman (approved):

PR #48, opened 2019-09-19
Offchain Spec (WIP)

lamafab:
Some adjustments based on your feedback. I'll follow up with the rest tomorrow. Just in case you wanted to take a look.

### DIFF
--- a/runtime-environment-spec/polkadot_re_spec.tm
+++ b/runtime-environment-spec/polkadot_re_spec.tm
@@ -3982,7 +3982,13 @@
 
   <subsection|Offchain Worker >
 
-  <todo|perhaps add an intro to this that what is an offchain worker>
+  Offchain Workers allow the execution of long-running and possibly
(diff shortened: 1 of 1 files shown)


tomaka:
For what it's worth, the HTTP methods have more precise documentation in this file: https://github.com/paritytech/substrate/blob/3f6cbc851883a7e7638e81ea43cf2bfc5e2a3736/core/primitives/src/offchain.rs

lamafab:
@tomaka Thanks, I already found that file, but a lot of the descriptions are Substrate specific and it's important to keep the spec Rust-independent. Certain areas (as marked in the file) must be expanded on.

tomaka:
To me what's missing from this PR, and what the file I link contains, is in which situations it is legal to call these functions, and in which situation they might return an error.

lamafab:
> To me what's missing from this PR, and what the file I link contains, is in which situations it is legal to call these functions, and in which situation they might return an error.

Okay, thanks for the feedback. I will adjust that.

PR #65, opened 2019-09-27
Integrate PDRE API tests + breaking changes fixes

lamafab:
**UPDATE (2019-10-11 12:42 CET**
The update includes:
- Generalized parameters for Julia tests
- Run each tester per function, not each function per tester (nested loops)
- Function output is now tested against a predefined list of "correct behavior"
- Julia support for verbosev and non-verbose output
- Parameter generator for calling Wasm functions. This reduced the lines of code by a huge margin

What's open:
- HTTP Offchain tests
- Splitting key signature tests into multiple sub-tests
- The predefined list of outputs should be expanded a little, since not all test functions return something.
- Add some documentation to the tests (and the internal functions/methods)

Take a look at the PR and give me a feedback. Let's merge this PR, the open adjustments and fixing the Substrate API breakage can be submitted in an upcoming PR (currently all tests pass). I'll focus on the libp2p network testing for now.

Output of non-verbose tests:
https://hastebin.com/cimeferitu.sql

Output of verbose tests:
https://hastebin.com/ludacokefi.sql

**UPDATE (2019-10-03 16:29 CET):**
The code was updated as requested. Data is now passed on (via Julia) like this:
`rust_tester pdre-api --function test_set_get_storage --input "Integrated,portal"`

Here is the example output of the tests: https://hastebin.com/ugozirinoq.sql

**Original post:**
As discussed... this integrates the PDRE API tests into `runtest.jl`, including the fixes for the breaking changes that occurred in the other tests. The RE tests are not fully done yet, but far enough.

Currently, there is no seperation between the logic and the data. IMHO, I'd leave it like this, since the data itself - unlike for the SCALE or Trie tests - is not that important here -> it's more about the RE having the specced interfaces and behavior correctly. I can change this, if you want (might take some time, though). The data can be directly seen in the `assert_eq!` macros. Where necessary, the data gets printed to `stdout`.

Output of running tests (notice the `PDRE API Test Suite` part):
```
$ ./run_tests.sh 
- Configuring done
-- Generating done
-- Build files have been written to: /home/anon/MyClones/polkadot-spec/build
[100%] running cargo
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
[100%] Built target rust_tester_target
[100%] Built target go_tester
[100%] Built target go_tester_all
[100%] running cargo
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
[100%] Built target rust_tester_target
[100%] Built target go_tester
[100%] Built target go_tester_all
Install the project...
-- Install configuration: ""
-- Up-to-date: bin/usr/local/bin/rust_tester
-- Installing: bin/usr/local/bin/go_tester
Test Summary:     | Pass  Total
Scale Codec Tests |    6      6
  1.005777 seconds (3.60 M allocations: 181.050 MiB, 2.42% gc time)
WARNING: replacing module CommonFixtures.
WARNING: replacing module ScaleCodecFixtures.
WARNING: replacing module StateTrieFixtures.
Test Summary:    | Pass  Total
State Trie Tests |   12     12
  0.709903 seconds (10.13 k allocations: 5.007 MiB)
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/rust_tester pdre-api`
***** PDRE API Test Suite *****
NOTE: not all tests display information, only those deemed necessary. If a test fails, this program panics.

Running crypto tests...
> testing ed25519 keys
  - public key 1: e735f8b3022b110c8d181f16e37bf04f59f334d770d5830fd9c9b213eaeb23b9
  - message: This is a message
  - signature: d82496fede03dd7a2ce0b9d7646f2382045684fae77a87972067ddec2ecac8255c992061e33d476a666a381911663a08ff9cfe5ee99f4c33bc65ba2176e89d0e
  - signature valid?: YES
  - public key 2: 14475a3d15623d4c0a92ef7ebab2b31a412929a9e25be88153095a6be7574811
  - all public keys : 14475a3d15623d4c0a92ef7ebab2b31a412929a9e25be88153095a6be7574811e735f8b3022b110c8d181f16e37bf04f59f334d770d5830fd9c9b213eaeb23b9
> testing sr25519 keys
  - public key 1: 74458252cc1bd2f482fd4cf94f82fd0211fd37970d2da4a7a964f3fd7fd6f72c
  - message: This is a message
  - signature: 2ed8e4057171914a9052a02dad17f8f7e0edff662f28f6d3f4fffd4c9a3d7422fc30073e1df72ba34cc7a07734186075a8d9cd868116b0026a48a3390577a887
  - signature valid?: YES
  - public key 2: d8a6d30076c291177ca6ab25caaa5328e6bae2ea1e9ba09f5400c7388f630007
  - all public keys : 74458252cc1bd2f482fd4cf94f82fd0211fd37970d2da4a7a964f3fd7fd6f72cd8a6d30076c291177ca6ab25caaa5328e6bae2ea1e9ba09f5400c7388f630007
> testing TwoX 64-bit hash
  - Input : TestInput
  - Output : 89d0f1c4908a48e3
> testing TwoX 128-bit hash
  - Input : TestInput
  - Output : 89d0f1c4908a48e3be56226b9fac6277
> testing TwoX 256-bit hash
  - Input : TestInput
  - Output : 89d0f1c4908a48e3be56226b9fac6277a97f2643e05a925bf8f01a359bd7f3c6
DONE - OK!
**********
Running network tests...
DONE - OK!
**********
Running storage tests...
DONE - OK!
Test Summary:             | Pass  Total
Runtime Environment Tests |    1      1
  1.975115 seconds (28.69 k allocations: 1.553 MiB)
  3.698028 seconds (3.64 M allocations: 187.622 MiB, 0.66% gc time)
```


test/pdre_api_tests.jl
```diff
@@ -0,0 +1,5 @@
+@testset "Runtime Environment Tests" begin
+    cd("test/testers/rust-tester/")
```
drskalman: hmm it should be written in a way that julia can easily run the test from other implementations and compare them. like https://github.com/w3f/polkadot-spec/blob/6d9d4b542e7a8bd8ab55fd2d95db1daf5745b928/test/scale_codec_tests.jl#L1 in particular: it need to 1. be able to run tests one by one so it can say which tests fails. 2. It should be able to store the stdout so it can compare it with other languages.

test/runtests.jl
```diff
@@ -3,6 +3,7 @@ using Test
 @time begin
   @time @testset "Scale Codec Tests" begin include("scale_codec_tests.jl") end
   @time @testset "State Trie Tests" begin include("state_trie_tests.jl") end
+  @time @testset "Runtime Environment Tests" begin include("pdre_api_tests.jl") end
```
drskalman: Perhapse "RE API Tests" is a better name (otherwise all the tests are runtime environment tests)

test/testers/rust-tester/src/cli.yaml
```diff
@@ -38,4 +38,6 @@ subcommands:
               short: i
               help: the file containing the key value data defining the state
               value_name: INPUT_VALUE
+    - pdre-api:
+        about: runtime environment API related tests
```
drskalman: maybe you can get the name of the function and its input(s) to be tested here

test/testers/rust-tester/src/pdre_api/crypto.rs
```diff
@@ -0,0 +1,179 @@
+use super::utils::CryptoApi;
+
+use substrate_primitives::hashing::{twox_128, twox_256, twox_64};
+
+pub fn test_blake2_128() {
+    let mut api = CryptoApi::new();
+
+    let data = String::from("TestInput");
```
drskalman: So better than test inputs are coming from https://github.com/w3f/polkadot-spec/blob/master/test/fixtures/test_fixtures.jl and will be fed to each implementation by julia so one can add and modify as much input as they want without touching the implementations.

drskalman: Same for all other functions.

test/testers/rust-tester/src/pdre_api/crypto.rs
```diff
@@ -0,0 +1,179 @@
...
+pub fn test_twox_64() {
+    let mut api = CryptoApi::new();
+
+    let data = String::from("TestInput");
+    let mut output = [0; 8];
+
+    api.rtm_ext_twox_64(data.as_bytes(), &mut output);
+    assert_eq!(twox_64(data.as_bytes()), output);
+
+    println!("> testing TwoX 64-bit hash");
```
drskalman: because other implementations need to exactly print the same thing I don't know if it is necessary to print anything beside output of the test. the julia has already given the input information and it has invoked the command line test with its name so it is aware of the input and the functionality being tested.

test/testers/rust-tester/src/pdre_api/crypto.rs
```diff
@@ -0,0 +1,179 @@
...
+    api.rtm_ext_sr25519_generate(keystore.as_bytes(), &[], &mut pubkey2);
+
+    // Get all public keys
+    let mut result_len: u32 = 0;
+    let all_pubkeys = api.rtm_ext_sr25519_public_keys(keystore.as_bytes(), &mut result_len);
+    assert_eq!(result_len, 65);
+
+    println!("> testing sr25519 keys");
+    println!("  - public key 1: {}", hex::encode(pubkey1));
+    println!("  - message: {}", msg);
```
drskalman: Digital signature are not canonical by definition so maybe we can gather them from various implementation and get them verified by all implementations.

test/testers/rust-tester/src/pdre_api/crypto.rs
```diff
@@ -0,0 +1,179 @@
...
+    println!("  - signature: {}", hex::encode(&signature[..]));
+    print!("  - signature valid?: ");
+    if verify == 0 {
+        println!("YES");
```
drskalman: "Good signature" is a better output upper or lower case.

test/testers/rust-tester/src/pdre_api/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+mod storage;
+mod utils;
+
+pub fn run_full_tests() {
+    println!("***** PDRE API Test Suite *****");
+    println!("NOTE: not all tests display information, only those deemed necessary. If a test fails, this program panics.");
+    println!("");
+
+    // crypto tests
+    println!("Running crypto tests...");
```
drskalman: So here I think something like this is expected https://github.com/w3f/polkadot-spec/blob/6d9d4b542e7a8bd8ab55fd2d95db1daf5745b928/test/testers/rust-tester/src/trie_tester.rs#L144 which decides which test to be run depending from the input given by the command line.

drskalman (changes requested):
Thanks  a lot! I haven't reviewed the rust code itself, just the code interface code which runs the test. I'll review the rust code in next review session.

test/pdre_api_tests.jl
lamafab: Done!

test/runtests.jl
lamafab: Done!

test/testers/rust-tester/src/cli.yaml
lamafab: Done!

test/testers/rust-tester/src/pdre_api/crypto.rs
lamafab: Done!

test/testers/rust-tester/src/pdre_api/crypto.rs
lamafab: Done!

test/testers/rust-tester/src/pdre_api/crypto.rs
lamafab: This one is trickier -> maybe this should be tested in the "big" testing phase where we tests all the Polkadot implementations (as discussed with Peter).

test/testers/rust-tester/src/pdre_api/crypto.rs
lamafab: Done!

test/testers/rust-tester/src/pdre_api/mod.rs
lamafab: Done!

test/fixtures/pdre_api_dataset.jl
```diff
@@ -0,0 +1,339 @@
+module PdreApiTestFixtures
+const cli_testers = [
```
drskalman, line 2: cli testers are going to be rust_tester, go_tester, cpp_tester, 
but only rust_tester for now

test/fixtures/pdre_api_dataset.jl
```diff
@@ -0,0 +1,339 @@
...
+  [
+    "build/bin/usr/local/bin/rust_tester", # cli
+    "pdre-api", # subcommand
+    "--function", # function arg
+    "--input" # input arg
```
drskalman: the commandline args should be same across cli_testers so they can have their own constant name instead of being in an arry.

test/pdre_api_tests.jl
```diff
@@ -0,0 +1,136 @@
+include("./fixtures/pdre_api_dataset.jl")
+using Test
+@testset "RE API Tests" begin
+    script_dir = @__DIR__
+    root_dir = script_dir * "/.."
+    manifest_dir = script_dir * "test/testers/rust-tester/Cargo.toml"
+    cd(root_dir)
+    for (cli, sub_cmd, func_arg, input_arg) in PdreApiTestFixtures.cli_testers
```
drskalman: so we don't need to loop on sub_cmd, ... only over clis. Moreover it make sense to loop over cli's as the most inner loop as we expect similar output of the same test across the cli testers.

test/pdre_api_tests.jl
```diff
@@ -0,0 +1,136 @@
...
+                # append input
+                cmd = string(cmd, " \"", value,"\"")
+
+                # Run
+                println(">> Running:", cmd)
+                output = read(`sh -c $cmd`, String)
+                if output != ""
+                    println(output)
+                end
+                @test true
```
drskalman, line 63: It makes sense to store the output from different cli testers in an array or so to do something with it. For example in case hashes to compare them and makes sure they are equal. In case of Signature maybe to run verify over them.

test/testers/rust-tester/src/pdre_api/crypto.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    let mut output = [0; 32];
+
+    api.rtm_ext_keccak_256(data, &mut output);
+
+    println!("Result: {}", hex::encode(output));
+}
+
+// Input: data
+pub fn test_ed25519(input: ParsedInput) {
```
drskalman, line 80: Maybe break this test into two one for generating the key and one for signing receiving the key as input.

test/testers/rust-tester/src/pdre_api/crypto.rs
```diff
@@ -0,0 +1,170 @@
...
+        println!("GOOD SIGNATURE");
+    } else {
+        println!("BAD SIGNATURE");
+    }
+    println!("Public key 2: {}", hex::encode(pubkey2));
+    println!("All public keys : {}", hex::encode(&all_pubkeys[1..]));
+}
+
+// Input: data
+pub fn test_sr25519(input: ParsedInput) {
```
drskalman, line 120: same here for sr25519

test/pdre_api_tests.jl
```diff
@@ -0,0 +1,136 @@
...
+            for value in PdreApiTestData.value_data
+                # create first part of the command
+                cmdparams = [cli, sub_cmd, func_arg, func, input_arg]
+                cmd = join(cmdparams, " ")
+
+                # append input
+                cmd = string(cmd, " \"", value,"\"")
+
+                # Run
+                println(">> Running:", cmd)
```
drskalman: I remember we said we make this conditional to something (either an cmd arg or a test fixture variable like "debug" or "verbose" or something)

test/testers/rust-tester/src/pdre_api/storage.rs
```diff
@@ -0,0 +1,382 @@
...
+    let res = api.rtm_ext_get_allocated_storage(key, &mut written_out);
+    assert_eq!(written_out, u32::max_value());
+    assert_eq!(res, [0u8; 0]);
+
+    // Set key/value
+    api.rtm_ext_set_storage(key, value);
+
+    // Get valid key
+    let mut written_out = 0;
+    let res = api.rtm_ext_get_allocated_storage(key, &mut written_out);
```
drskalman, line 26: I'd either print or return the get valued so the julia can do something with it (so we have a standard way of failing a test).

test/testers/rust-tester/src/pdre_api/storage.rs
```diff
@@ -0,0 +1,382 @@
...
+        println!("Key `{}` was deleted", str(key2));
+    } else {
+        assert_eq!(written_out, value2.len() as u32);
+        assert_eq!(res, value2);
+        println!("Key `{}` remains", str(key2));
+    }
+}
+
+// Input: child1, child2, key, value
+pub fn test_set_get_child_storage(input: ParsedInput) {
```
drskalman: I'm thinking that it makes sense to move the chid storage tests to another file.

test/testers/rust-tester/src/pdre_api/utils/crypto.rs
```diff
@@ -0,0 +1,513 @@
...
+                    ])
+                },
+                |_, memory| {
+                    output.copy_from_slice(
+                        memory
+                            .get(*ptr_holder.borrow(), 32)
+                            .map_err(|_| Error::Runtime)?
+                            .as_slice(),
+                    );
+                    Ok(Some(()))
```
drskalman: This part of the code is seems to be repeating with minor changes. I'm sure you can add a middle function(s) minimize the repetition. Deos this lambdish \lambda(aloc) and result \lambda(.memory) functions need to defined in line? Can't they be an actual functions so you don't need to redefine them at each call of call_with_costume_sginature?

test/testers/rust-tester/src/pdre_api/utils/storage.rs
```diff
@@ -0,0 +1,441 @@
...
+                        *written_out = u32::from_le_bytes(
+                            memory
+                                .get(*ptr_holder.borrow(), 4)
+                                .unwrap()
+                                .as_slice()
+                                .split_at(4)
+                                .0
+                                .try_into()
+                                .unwrap(),
+                        );
```
drskalman: Same repetition mitigation applies here.

test/testers/rust-tester/src/pdre_api/utils/storage.rs
```diff
@@ -0,0 +1,441 @@
...
+                            .map(Some)
+                    } else {
+                        Ok(None)
+                    }
+                },
+            )
+            .unwrap()
+    }
+
+    pub fn rtm_ext_get_allocated_child_storage(
```
drskalman: Here also I'd cut out the child storage related routines into another file.

test/pdre_api_tests.jl
lamafab: Just to clarify; the hierarchy of the loops should be changed from `cli > tests` to `tests > cli`?

Therefore (pseudo-code):
```
loop crypto_tests {
    loop cli {
        ...
    }
}
loop network_tests {
    loop cli {
        ...
    }
}
...
```

drskalman: I think so, I think Cli should be the most inner one
```
loop crypto_tests
  loop blake_256 inputs
      loop over rust_tester, go_tester, ...
```
 So we just compare outputs that single input for different cli_testers's when each cli_testers (rust, go, c++) finished and gave the output for that one input. Like here:  

https://github.com/w3f/polkadot-spec/blob/f4fa2f0ccb3edaee0f252ff52231709bf780c063/test/scale_codec_tests.jl#L8

lamafab: Ok, will be adjusted.

test/testers/rust-tester/src/pdre_api/crypto.rs
lamafab: While implementing this I ran into a problem; the entire process from generating, signing and verifying must be done in one function, since the storage is not persistent (the RE does not support importing keys). Honestly I'd leave it as it is and just print the values. Those keys and signatures are not deterministic anyway.

drskalman: Whatever you need to store print it out (the key, the signature),to stdout then julia can read it and feed it to the next function as a command line argument

PR #71, opened 2019-10-23
Generalise & document dependency on algorithmacs-style.ts

romanb:
The algorithmacs-style.ts file is currently referenced with a hardcoded path of a specific system (presumably the main authors'). It is now assumed to be in the user's personal style package directory. The file is also added to this repository, just in case the referenced original on gitlab disappears.

I couldn't make TeXmacs use a relative path to the style file (I never used TeXmacs before), which would have been my preference. Requiring it to be added to the personal style package directory seemed the next best option.


drskalman:
I have cloned the repo to w3f github: https://github.com/w3f/algorithmacs so perhapse better if we link it this repo.
Also I'm wondering if it is better that we add it as a submodule and and point to it with relative address?

romanb:
> I have cloned the repo to w3f github: https://github.com/w3f/algorithmacs so perhapse better if we link it this repo.

Of course, I didn't see that, thanks for the pointer. I updated the PR, also to better not include a copy of the file but rather just point to the other repository in the README, so changes to the style file don't require an update here (users /  editors just need to pull the updated style file from the other repository).

> Also I'm wondering if it is better that we add it as a submodule and and point to it with relative address?

That may be preferable but as I mentioned in the PR description, I didn't find a way to get a relative path to work. I personally think just cloning the other repository and setting up a (sym)link is convenient enough though.

drskalman:
Reading Texmacs guide, I realized that if you copy the style file in ~/.TeXmacs/packages/algorithmacs-style.ts and change this in the file:
```
-<style|<tuple|tmbook|std-latex|/home/klaymen/doc/code/algorithmacs/algorithmacs-style.ts>>
+<style|<tuple|tmbook|std-latex|algorithmacs-style>>
```
It works like a charm. Could you test, confirm and update the README and the pull request?
Thanks a lot.

romanb:
> Reading Texmacs guide, I realized that if you copy the style file in ~/.TeXmacs/packages/algorithmacs-style.ts and change this in the file:
> 
> ```
> -<style|<tuple|tmbook|std-latex|/home/klaymen/doc/code/algorithmacs/algorithmacs-style.ts>>
> +<style|<tuple|tmbook|std-latex|algorithmacs-style>>
> ```
> 
> It works like a charm. Could you test, confirm and update the README and the pull request?
> Thanks a lot.

That is exactly what this PR has been doing from the start. I only included the `.ts` file earlier for convenience because I didn't know about the other repository. Please take a look at the diff.

lamafab (approved):
Just pulled it, looks good.

Thank you @romanb. Can I ask; are you employed by an organization that works with/on Polkadot? Or what's your relationship with this project here?

romanb:
@lamafab Thanks. I currently work at @paritytech and am studying the networking protocols used in substrate/polkadot more closely.

PR #74, opened 2019-11-06
Spec key types used by Polkadot

lamafab:
**Update 2019-11-19:**
The descriptions of the relevant key types were adjusted. TODO: Spec how the Controller and Session keys are certified in the network.

**Original description:**
The added information about key types, as discussed. Since only Session keys are being used by the Runtime, should all those key types even be mentioned? Maybe leave those out completely?

I think it's relevant information for understanding the concept. Feedback appreciated.

Printscreen:
image


burdges:
I guess this is the current keys since many of the session key types must change.

lamafab:
> I guess this is the current keys since many of the session key types must change.

@burdges Can you elaborate on that part?

PR #79, opened 2019-11-09
Add additional dependencies in circleci config

lamafab:
Should fix the Circle CI builds.


lamafab:
Okay, it's finally fixed now. Merging.

PR #80, opened 2019-11-19
Spec requested runtime entries

lamafab:
Specs the [requested](https://hackmd.io/qP__ItARSKy5TtSnpSVL0A?view) runtime entries:

```
BLOCKING. Runtime entries:
BlockBuilder::inherent_extrinsics,
Core::initialize_block,
BlockBuilder::apply_extrinsic,
BlockBuilder::finalise_block
```

Please note that certain data types will be [changed upstream](https://github.com/paritytech/substrate/pull/4134).

Feedback welcome

**Screenshots**:
image
image
image


runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -5245,17 +5263,37 @@
 
   <strong|Return>:
 
-  A Boolean value indicates if the execution was successful.
+  <\itemize-dot>
+    <item>A Boolean value indicates if the execution was successful.
+  </itemize-dot>
 
   <subsection|<verbatim|Core_initialise_block>>
```
rphmeier: should be `initialize`

rphmeier:
I notice some inconsistency w.r.t. saying whether arguments or return types are SCALE-encoded or not. Either all arguments and return types should be explicitly labelled as SCALE, or there should be a note above clarifying that all argument lists and return types are SCALE (going a bit deeper into how multiple arguments are interpreted) and then none of the APIs themselves should repeat that.

The half-and-half here is just confusing, since it implies that e.g. the header passed to `initialize_block` is not SCALE

lamafab:
@rphmeier Okay, I'll adjust that.

Issue #104, opened 2020-01-08
Improve the definition of core_initialize_block in the appendix


drskalman:
- It needs to specifies which block's header is needed (parent or itself, it is probably itself).
   - It needs to specifies that it is called during production as well, because the appendix only it says it is called during block execution but it doesn't mention that it is called during block building as Algorithm 5.7 shows.
   - Verify that it actually being called on full client during received block execution.
   - Algorithm 5.7 building block should mention the complete name of the RT entries.

lamafab:
Current progress:

- [x] It needs to specifies which block's header is needed (parent or itself, it is probably itself).
- [x] It needs to specifies that it is called during production as well, because the appendix only it says it is called during block execution but it doesn't mention that it is called during block building as Algorithm 5.7 shows.
- [x] Verify that it actually being called on full client during received block execution.
- [x] Algorithm 5.7 building block should mention the complete name of the RT entries.

PR #106, opened 2020-01-09
Expand description of `Core_initialize_block` and `Core_execute_block`

lamafab:
Expands the descriptions of the Runtime entries requested in #104. If there is any phrasing you would change let me know.

Printscreen 1 (updated):
image

Printscreen 2 (updated):
image


lamafab:
I think `Core_inherent_extrinsics` should be renamed to `BlockBuilder_inherents_extrinsics`... I'll change that.

EDIT: Done.

PR #119, opened 2020-01-30
Spec the new PDRE APIs

lamafab:
This PR includes the specification for the Storage, Child Storage, Crypto, Hashing and Offchain APIs. Additional types have been declared.

**Todo**
- [x] I'm waiting for further clarification for certain types passed to Child Storage APIs (those are marked in the document).
- [x] Two Crypto (recovery) functions and the Trie functions have been skipped for now, since those are slightly more complex to spec. I will update those once I get back to writing the tests and can actually experiment with the values/outputs (as discussed, I'll do the "ForcedChange" and the Transaction Weights first). The handful of other APIs that we have left out before - such as Allocator and Logging/printing - I will spec in the upcoming PR, too. Then we can finally finish this chapter.


drskalman:
Generally looks good. Few points:
 
 - [x] I change the runtime pointer to pointer-size type which is more explicit and accurate of what is being store to i64. Could you update it everywhere?
 
 - [x] Could the function names in the section title be verbatimize so it represents the fact that they are actually the exact function names used in the wasm?
 
 - [x] I think the Version 1 prototype is more approperiate to be a \subsubsection rather than just bold/strong font as different version might have different argunemnt.
   
I applied all these changes to E.1.1. ext_storage_set, if you could apply them to all other functions that would be very good.

ext_storage_set

drskalman:
Could you clarify all the encoding (perhapse in the intro) for option, result and the integers which you don't explicitly specify in the function definition?

PR #142, opened 2020-03-06
Spec Changes Trie

### LINKED ISSUE #129: Spec Storage changes trie
Storage change trie root is part of digest 
   https://github.com/paritytech/substrate/blob/master/test-utils/runtime/src/system.rs#L217
   
   And we need to spec how it is generated.

lamafab:
Requested spec for #129.

@drskalman @FlorianFranzen  Putting the entire behavior and steps of building the Changes Trie into words is kind of hard, so please leave feedback if something isn't clear or needs to be clarified/expanded.

UPDATE (2020-03-10) This was adjusted: ~~Also, in the code there's this special case of `NO_EXTRINSIC_INDEX`, I'll follow up with that (either in this PR or a new one).~~

UPDATE (2020-04-15) Big restructuring based on the feedback I've received. Clarified some sections and added examples. Current Screenshot:

image


runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+    changes tries mappings> track any storage keys from child storages and
+    their corresponding Child Changes Trie. As described in Definition
+    <reference|defn-storage-key-to-extrinsics>, changes inside child storages
+    are inserted into their own Trie. The changed key is inserted into the
+    main Trie, where it's value is the Merkle proof as desribed in section
+    <reference|sect-merkl-proof> of the corresponding Child Storage Trie,
+    which stores the mappings between the storage key and exstrinsics.
+  </definition>
+
+  The Trie itself is not part of the block, but a separately maintained
```
Harrm: Do you mean the main trie or the changes trie here? Please clarify.

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+
+    \;
+  <|big-table>
+    <label|table-changes-trie-key-types>Varying datatype of possible mappings
+    in the Changes Trie
+  </big-table>
+
+  <strong|Note>: the key used for the Changes Trie must be the <em|varying
+  datatype>, not the individual, appended <em|KeyIndex>. Unlike the default
+  encoding for varying data types, this structure starts it's indexing at
```
Harrm:
```suggestion
  encoding for varying data types, this structure starts its indexing at
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+  <reference|defn-storage-key-to-blocks> and
+  <reference|defn-storage-key-to-child-tries>.
+
+  <\big-table>
+    <tabular|<tformat|<cwith|2|2|1|-1|cell-bborder|0ln>|<cwith|3|3|1|-1|cell-tborder|1ln>|<cwith|3|3|1|-1|cell-bborder|0ln>|<cwith|4|4|1|-1|cell-tborder|1ln>|<cwith|4|4|1|-1|cell-bborder|1ln>|<cwith|4|4|1|1|cell-lborder|0ln>|<cwith|4|4|3|3|cell-rborder|0ln>|<cwith|1|1|1|-1|cell-tborder|1ln>|<cwith|1|1|1|-1|cell-bborder|1ln>|<cwith|2|2|1|-1|cell-tborder|1ln>|<cwith|1|1|1|1|cell-lborder|0ln>|<cwith|1|1|3|3|cell-rborder|0ln>|<table|<row|<cell|<strong|Id>>|<cell|<strong|Description>>|<cell|<strong|Appended>>>|<row|<cell|1>|<cell|Mapping
+    between storage key and extrinsics (<reference|defn-storage-key-to-extrinsics>)>|<cell|<em|KeyIndex>
+    (<reference|defn-key-index>)>>|<row|<cell|2>|<cell|Mapping between
+    storage key and blocks (<reference|defn-storage-key-to-blocks>)>|<cell|<em|KeyIndex>
+    (<reference|defn-key-index>)>>|<row|<cell|3>|<cell|Mapping between
+    storage key and child change Trie (<reference|defn-storage-key-to-child-tries>)>|<cell|<em|KeyIndex>
```
Harrm:
```suggestion
    storage key and child changes Trie (<reference|defn-storage-key-to-child-tries>)>|<cell|<em|KeyIndex>
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+  datatype>, not the individual, appended <em|KeyIndex>. Unlike the default
+  encoding for varying data types, this structure starts it's indexing at
+  <verbatim|1>.
+
+  <\definition>
+    <label|defn-key-index>The <strong|Key Index> is a tuple containing the
+    following information:
+
+    <\big-table|<tabular|<tformat|<cwith|2|2|1|1|cell-lborder|0ln>|<cwith|2|2|3|3|cell-rborder|0ln>|<cwith|3|3|1|-1|cell-tborder|1ln>|<cwith|2|2|1|-1|cell-bborder|1ln>|<cwith|3|3|1|-1|cell-bborder|1ln>|<cwith|3|3|1|1|cell-lborder|0ln>|<cwith|3|3|3|3|cell-rborder|0ln>|<cwith|1|1|1|-1|cell-tborder|1ln>|<cwith|1|1|1|-1|cell-bborder|1ln>|<cwith|2|2|1|-1|cell-tborder|1ln>|<cwith|1|1|1|1|cell-lborder|0ln>|<cwith|1|1|3|3|cell-rborder|0ln>|<table|<row|<cell|<strong|Name>>|<cell|<strong|Description>>|<cell|<strong|Type>>>|<row|<cell|Block>|<cell|Block
+    at which this key has been inserted in the trie>|<cell|Unsigned 32-bit
```
Harrm: Please clarify here what exactly is this 32bit number (e. g. it's not completely clear if it's the hash of the header or the state root hash).

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+    encoded array containing the indices of the extrinsics that caused any
+    changes to the key. The indices are unsigned 32-bit integers and it's
+    values depend on the order in which each extrinsic is included in the
+    block. Additionally, a separate Changes Trie is created for each child
+    storage where changes occure.
+
+    \;
+
+    Keys that were modified outside of extrinsics, such as changes made by
+    the Runtime directly, have a <verbatim|0xffffffff> value as it's indice.
```
Harrm:
```suggestion
    the Runtime directly, have a <verbatim|0xffffffff> value as their index.
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+
+    Those entries are generated when the Runtime calls
+    <verbatim|ext_storage_changes_root> as described in section
+    <reference|sect-ext-storage-changes-root>.
+  </definition>
+
+  <\definition>
+    <label|defn-storage-key-to-blocks>The <strong|storage key to blocks
+    mappings> track any changes which occured in a certain range of blocks.
+    The key is inserted into the Trie, where it's value is a SCALE encoded
```
Harrm:
```suggestion
    The key is inserted into the Trie, where its value is a SCALE encoded
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+    array containing block numbers where extrinsics caused any changes to the
+    key. The block numbers are represented as unsigned 32-bit integers.
+
+    \;
+
+    Those entries are generated when the Runtime calls
+    <verbatim|ext_storage_changes_root> as described in section
+    <reference|sect-ext-storage-changes-root>, but unlike the other mappings,
+    they are not generated on each block. The Runtime sets the key
+    <verbatim|:changes_trie> to storage, where it's value is made out of two
```
Harrm:
```suggestion
    <verbatim|:changes_trie> to storage, where its value is made out of two
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+    (where this kind of mappings were created) + 1 to the current block.
+  </definition>
+
+  <\definition>
+    <label|defn-storage-key-to-child-tries>The <strong|storage key to child
+    changes tries mappings> track any storage keys from child storages and
+    their corresponding Child Changes Trie. As described in Definition
+    <reference|defn-storage-key-to-extrinsics>, changes inside child storages
+    are inserted into their own Trie. The changed key is inserted into the
+    main Trie, where it's value is the Merkle proof as desribed in section
```
Harrm:
```suggestion
    main Trie, where its value is the Merkle proof as desribed in section
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+    <reference|sect-merkl-proof> of the corresponding Child Storage Trie,
+    which stores the mappings between the storage key and exstrinsics.
```
Harrm:
```suggestion
    which stores the mappings between the storage key and extrinsics.
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,131 @@
...
+  </definition>
+
+  The Trie itself is not part of the block, but a separately maintained
+  database by the Polkadot node. The Merkle proof of the Change Trie must be
```
Harrm:
```suggestion
  database by the Polkadot node. The Merkle proof of the Changes Trie must be
```

runtime-environment-spec/polkadot_re_spec.tm
lamafab: The Changes Trie. Spec adjusted.

runtime-environment-spec/polkadot_re_spec.tm
lamafab: It's just the block number. Spec adjusted.

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,132 @@
...
+      Key structure inserted in the Changes Trie
+    </big-table>
+  </definition>
+
+  <\definition>
+    <label|defn-storage-key-to-extrinsics>The <strong|storage key to
+    extrinsics mappings> track any changes which occure in an individual
+    block. The key is inserted into the Trie, where its value is a SCALE
+    encoded array containing the indices of the extrinsics that caused any
+    changes to the key. The indices are unsigned 32-bit integers and it's
```
Harrm:
```suggestion
    changes to the key. The indices are unsigned 32-bit integers and its
```

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,132 @@
...
+    storage where changes occure.
+
+    \;
+
+    Keys that were modified outside of extrinsics, such as changes made by
+    the Runtime directly, have a <verbatim|0xffffffff> value as its indice.
+
+    \;
+
+    Those entries are generated when the Runtime calls
```
Harrm: It's not quite clear to me what is meant by generation of the entries. I may try to guess or ask you, but as it's not quite clear now, maybe you might clarify it a bit.

runtime-environment-spec/polkadot_re_spec.tm
```diff
@@ -1937,7 +1937,137 @@
...
+    number at which this key has been inserted in the trie>|<cell|Unsigned
+    32-bit integer>>|<row|<cell|Key>|<cell|The changed key>|<cell|Byte
+    array>>>>>>
+      Key structure inserted in the Changes Trie
+    </big-table>
+  </definition>
+
+  <\definition>
+    <label|defn-storage-key-to-extrinsics>The <strong|storage key to
+    extrinsics mappings> track any changes which occure in an individual
```
Harrm: occur/occured?

lamafab: -> Any changes that happened before `ext_storages_changes_root` gets called. I'm expanding a section in the storage state transition which explains how the changes get tracked, I'll push it today.

Harrm: I mean, there's a typo, but still thank you.

FlorianFranzen (changes requested):
Looks good. I rewrote the paragraph before table 2.1 a little bit and added other minor fixes (like updating the TOC). The changes are available in a separate pull request to this branch, see #160.

lamafab:
> Looks good. I rewrote the paragraph before table 2.1 a little bit and added other minor fixes (like updating the TOC). The changes are available in a separate pull request to this branch: w3f/polkadot-spec/pull#160

Yes, I saw that. I'll look into it.

drskalman:
@lamafab could you move it to c02-state.tm?

polkadot-host-spec/c02-state.tm
```diff
@@ -476,42 +476,193 @@
...
+  described in section <reference|sect-ext-storage-changes-root>. The
+  Polkadot Host requires a mechanism to identify the changes created by a
+  block, as mentioned in section <reference|sect-state-storage>.
+
   \;
+
+  The Changes Trie tracks changes in three different types of mappings (or
+  search indices) which are differentiated by their key as defined in table
+  <reference|table-changes-trie-key-types>. inside the trie. While the
+  overall structure of each keys is almost identical, each mapping has it own
```
drskalman: What is "almost" here? they seems identical to me.

polkadot-host-spec/c02-state.tm
```diff
@@ -476,42 +476,193 @@
...
+  prefix and could therefore be considered a <with|font-shape|italic|varying
+  datatype> (as defined in <reference|defn-varrying-data-type>). In contrast
+  to a <with|font-shape|italic|varying datatype> however, the type
+  identifiers do not describes the type of the key following it, but the
+  corresponding value of that key. The format and use of these values are
```
drskalman: why it is not just simply a tuple type then? and so if it, I would define it as " the tracking change type is a SCALE Encoded tuple
(id, keyindex) in which id is in table blah and key index is blah.

polkadot-host-spec/c02-state.tm
```diff
@@ -476,42 +476,193 @@
...
+  a datastructure maintained by the Polkadot node which tracks storage
+  changes created by each block. The primary method of generating the Changes
+  Trie is by calling the <verbatim|ext_storage_changes_root> Host API as
+  described in section <reference|sect-ext-storage-changes-root>. The
+  Polkadot Host requires a mechanism to identify the changes created by a
+  block, as mentioned in section <reference|sect-state-storage>.
+
   \;
+
+  The Changes Trie tracks changes in three different types of mappings (or
```
drskalman: When we talk about "Changes Trie" I think at some point we need to formally define it as a trie. Is it radix-16 trie for example. Are the nodes have exactly the same structure as the state trie etc. So maybe turning this paragraph into a definition saying what type of tree it is and then the key type and value type etc. 
"Track changes in three different types" what does this algorithmically mean? Like we store the key that is change in a trie node? it is not really clearly at least from the first read.

polkadot-host-spec/c02-state.tm
```diff
@@ -476,42 +476,193 @@
...
+    following information:
+
+    <\big-table|<tabular|<tformat|<cwith|2|2|1|1|cell-lborder|0ln>|<cwith|2|2|3|3|cell-rborder|0ln>|<cwith|3|3|1|-1|cell-tborder|1ln>|<cwith|2|2|1|-1|cell-bborder|1ln>|<cwith|3|3|1|-1|cell-bborder|1ln>|<cwith|3|3|1|1|cell-lborder|0ln>|<cwith|3|3|3|3|cell-rborder|0ln>|<cwith|1|1|1|-1|cell-tborder|1ln>|<cwith|1|1|1|-1|cell-bborder|1ln>|<cwith|2|2|1|-1|cell-tborder|1ln>|<cwith|1|1|1|1|cell-lborder|0ln>|<cwith|1|1|3|3|cell-rborder|0ln>|<table|<row|<cell|<strong|Name>>|<cell|<strong|Description>>|<cell|<strong|Type>>>|<row|<cell|Block>|<cell|Block
+    number at which this key has been inserted in the trie>|<cell|Unsigned
+    32-bit integer>>|<row|<cell|Key>|<cell|The changed key>|<cell|Byte
+    array>>>>>>
+      Key structure inserted in the Changes Trie
+    </big-table>
+  </definition>
+
```
drskalman: For all these 3 definitions I would define a formal function (give it a name) define its  domain and range type and it is criteria (similar the childrenbitmap for example). It will be less confusing.

polkadot-host-spec/c02-state.tm
```diff
@@ -476,42 +476,193 @@
...
+  <\definition>
+    <label|defn-storage-key-to-blocks>The <strong|storage key to blocks
+    mappings> track any changes which occured in a certain range of blocks.
+    Its value is a SCALE encoded array containing block numbers where
+    extrinsics caused any changes to its key. The block numbers are
+    represented as unsigned 32-bit integers.
+
+    \;
+
+    The Changes Trie is generated when the Runtime calls
```
drskalman: It seems that stuff here are so is specifically about ext-storage-changes-root and is about how to basically call that API. Why shouldn't this part be moved into sect-ext-storage-changes-root ?

polkadot-host-spec/c02-state.tm
```diff
@@ -476,42 +476,193 @@
...
+    <label|defn-storage-key-to-child-tries>The <strong|storage key to child
+    changes tries mappings> track any storage keys from child storages and
+    their corresponding Child Changes Trie. As described in Definition
+    <reference|defn-storage-key-to-extrinsics>, changes inside child storages
+    are inserted into their own Trie. The changed key is inserted into the
+    main Trie, where its value is the Merkle proof as desribed in section
+    <reference|sect-merkl-proof> of the corresponding Child Storage Trie,
+    which stores the mappings between the storage key and extrinsics.
+  </definition>
+
```
drskalman: We don't really talk about blocks before chapter 3. Shouldn't the following paragraph be moved to to where we spec the header of the block and then refer back to this section?

drskalman (changes requested):

polkadot-host-spec/c02-state.tm
lamafab: Well, that enum ("varying data type"), which is the key, indicates the type of the *value*, not the key itself.

We define "varying data type" as:
image

In this case it's:
image

I guess we can just leave it at "it's a varying data type" and just explicitly say that it refers to the type of the other value, not the key itself.

polkadot-host-spec/c02-state.tm
lamafab: See [here](https://github.com/w3f/polkadot-spec/pull/142#discussion_r404836931).

polkadot-host-spec/c02-state.tm
lamafab: I'll reference the radix-16 trie in the Changes Trie section.

> "Track changes in three different types" what does this algorithmically mean?

Depending on the enum, different value types gets inserted into that trie. That type is identified by the enum index. The mappings are:

```
<changed-key> --> <array-of-extrinsic-indexes>
<changed-key> --> <array-of-block-numbers>
<changed-key> --> <array-of-child-changes-tries>
```

I'll change the "tracks changes *in*..." to "tracks changes *of*" this as:

> The Changes Trie tracks changes of three different types of mappings (or search indices) which are differentiated by their key as defined in table 2.1.

I'm also replacing "Followed by" with "Value":
image

polkadot-host-spec/c02-state.tm
lamafab: This was moved to the `ext_storage_changes_root` section.

polkadot-host-spec/c02-state.tm
lamafab: As discussed, the Changes Trie section was moved to chapter 3.

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,198 @@
...
+    Child Storage Trie.
+  </definition>
+
+  The Changes Trie itself is not part of the block, but a separately
+  maintained database by the Polkadot node. The Merkle proof of the Changes
+  Trie must be inlcuded in the block digest as described in Definition
+  <reference|defn-digest> and gets calculated as described in section
+  <reference|sect-merkl-proof>. The root calculation uses entries of a
+  per-block basis (including block mappings when conditions apply on that
+  block), not the whole changes spanning multiple blocks.
```
Harrm: The meaning of this is not quite clear. For a block where block mappings are generated the changes trie must include entries related to any extrinsic changed in the tracked range of blocks, right? What does per-block basis mean in this case?

polkadot-host-spec/ae-hostapi.tm
```diff
@@ -204,10 +204,38 @@
...
+  Unlike extrinsic mappings, block mappings are not created for each block.
+  The Runtime sets the key <verbatim|:changes_trie> in storage with its value
+  being a tuple of two 32-bit integers:
+
+  <\itemize-dot>
+    <item><verbatim|digest_interval> - The interval (in blocks) at which
+    block mappings are created. Block mappings are not created when this is
+    less or equal to <verbatim|1>.
+
+    <item><verbatim|digest_levels> - Maximal number of \Plevels\Q in the
```
Harrm: I don't really understand what hierarchy is meant here. Does that mean that if, for example, I have interval of 3 and levels of 4, I have to add to the changes trie of every 3rd block the entries about  changes made in past 3 blocks, to every 9th block changes made in past 9 blocks, to every 27th block for past 27 blocks and so on to 3^4? So I have to store the information about what keys were changed by which blocks for 81 blocks?

polkadot-host-spec/c02-state.tm
```diff
@@ -16,6 +16,14 @@
   limits on their length. The limit is imposed by the encoding algorithms to
   store the key and the value in the storage trie.
 
+  \;
+
+  The state storage requires a mechanism to identify any changes made by each
+  block, in case non-finalized blocks have to be reverted or a Changes Trie
+  as described in section <reference|sect-general-tree-structure> needs to be
```
Harrm: Could you clarify  the meaning of "identify any changes made by each block"? Does that mean that there should be a mechanism to get the list of keys changed by a block for any block in the block tree?

polkadot-host-spec/ae-hostapi.tm
```diff
@@ -204,10 +204,38 @@
...
   </itemize>
 
-  <subsection|<verbatim|ext_storage_changes_root>>
+  <subsection|<verbatim|ext_storage_changes_root>><label|sect-ext-storage-changes-root>
 
-  Commits all existing operations and gets the resulting change root. The
-  parent hash is a SCALE encoded change root.
+  Commits all existing operations and gets the resulting change root as
+  described in section <reference|sect-changes-trie>. The parent hash is a
+  SCALE encoded change root.
```
Harrm: Sorry, not quite clear what is a change root.

polkadot-host-spec/ae-hostapi.tm
FlorianFranzen: Well, the nice thing about the way this spacing works is

FlorianFranzen: You only need to keep a handful of digest tries around. E.g. for `interval = 3` and `level = 4` the you would have a try over 3, 9, 27, 81 blocks. Whenever one of these intervals has passed, you add the current trie for this interval to the merkel prove. Then the trie of that interval is reset (I assume they are also added to a local database somewhere). If you want to save on computation you could update the higher intervals only when the interval just below it is reset. Because the intervals contain multiples of each other, you can use the already consolidated tries form the lower levels instead of adding each single transaction to each interval trie every single time.

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+    between storage key and extrinsics (section
+    <reference|sect-changes-trie-extrinsics-pairs>)>|<cell|<math|<around*|{|e<rsub|i>,\<ldots\>,e<rsub|k>|}>>>>|<row|<cell|>|<cell|<text-dots>
+    where <math|e<rsub|i> refers to the >indice of the extrinsic within the
+    block>|<cell|>>|<row|<cell|2>|<cell|Pair between storage key and block
+    numbers (section <reference|sect-changes-trie-block-pairs>)>|<cell|<math|<around*|{|H<rsub|i><around*|(|B<rsub|k>|)>,\<ldots\>,H<rsub|i><around*|(|B<rsub|m>|)>|}>>>>|<row|<cell|3>|<cell|Pair
+    between storage key and Child Changes Trie (section
+    <reference|sect-changes-trie-child-trie-pair>)>|<cell|<math|H<rsub|r><around*|(|<text|<name|Child-Changes-Trie>>|)>>>>>>>
```
drskalman: - is "Child-Changes-Trie" is a type? if we are storing the root here then maybe it is better to call it the root of ...
- You might want to remind the reader what is H_r.

drskalman: - is "Child-Changes-Trie" is a type? if we are storing the root here then maybe it is better to call it the root of ...

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+  structure starts its indexing at <verbatim|1>.
+
+  \;
+
+  The Changes Trie itself is not part of the block, but a separately
+  maintained database by the Polkadot Host. The Merkle proof of the Changes
+  Trie must be included in the block digest as described in Definition
+  <reference|defn-digest> and gets calculated as described in section
+  <reference|sect-merkl-proof>. The root calculation only considers pairs
+  which were generated on the individual block and does not consider pairs
```
drskalman: - I wouldn't call them pairs. I'd call them "nodes".
- It is not clear what is happening here. I'm not sure why we need to have word "consider". It is better state that Host make a different change trie for each block whose nodes only consists of the keys that are changed in that block. (even if substrate is keeping one change trie and do some magic to keep other nodes out of calculation).

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+  to be used by \Pproof servers\Q, where its implementation and behavior has
+  not been fully defined yet. This is considered future-reserved>
+
+  \;
+
+  As clarified in the individual sections of each type, not all of those
+  types get generated on every block. But if conditions apply, all those
+  different types of pairs get inserted into the same Changes Trie, therefore
+  only one Changes Trie Root gets generated for each block.
+
```
drskalman: This confusing. I'm not sure if it should be here. you haven't given an algorithm to generate the trie yet. So as a reader, I assume that something will go in the trie not everything in the universe goes into the trie. So I don't see the reason for this warning. I'd just say "We describe the algorithm to generate the change trie on each block in section blah".

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+  <reference|table-changes-trie-key-types>.<htab|5mm>
+
+  <\big-table>
+    <tabular|<tformat|<cwith|2|2|1|-1|cell-bborder|0ln>|<cwith|4|4|1|-1|cell-tborder|1ln>|<cwith|4|4|1|-1|cell-bborder|0ln>|<cwith|5|5|1|-1|cell-tborder|1ln>|<cwith|5|5|1|-1|cell-bborder|1ln>|<cwith|5|5|1|1|cell-lborder|0ln>|<cwith|5|5|3|3|cell-rborder|0ln>|<cwith|1|1|1|-1|cell-tborder|1ln>|<cwith|1|1|1|-1|cell-bborder|1ln>|<cwith|2|2|1|-1|cell-tborder|1ln>|<cwith|1|1|1|1|cell-lborder|0ln>|<cwith|1|1|3|3|cell-rborder|0ln>|<table|<row|<cell|<strong|Type>>|<cell|<strong|Description>>|<cell|<strong|Value>>>|<row|<cell|1>|<cell|Pair
+    between storage key and extrinsics (section
+    <reference|sect-changes-trie-extrinsics-pairs>)>|<cell|<math|<around*|{|e<rsub|i>,\<ldots\>,e<rsub|k>|}>>>>|<row|<cell|>|<cell|<text-dots>
+    where <math|e<rsub|i> refers to the >indice of the extrinsic within the
+    block>|<cell|>>|<row|<cell|2>|<cell|Pair between storage key and block
+    numbers (section <reference|sect-changes-trie-block-pairs>)>|<cell|<math|<around*|{|H<rsub|i><around*|(|B<rsub|k>|)>,\<ldots\>,H<rsub|i><around*|(|B<rsub|m>|)>|}>>>>|<row|<cell|3>|<cell|Pair
```
drskalman: Reading these my understanding is that type 1 and 3 nodes are computed and inserted into the change trie of each block, while type 2 nodes are inserted in a trie that is generate with specific request of runtime. If that is the case then I wouldn't mention type 2 here. I move it to API section where it generates change tries which also have type 2 nodes.

polkadot-host-spec/c03-transition.tm
drskalman: - again if we are generating try always without type 2 then I wouldn't mention type 2 in this section.

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+  not been fully defined yet. This is considered future-reserved>
+
+  \;
+
+  As clarified in the individual sections of each type, not all of those
+  types get generated on every block. But if conditions apply, all those
+  different types of pairs get inserted into the same Changes Trie, therefore
+  only one Changes Trie Root gets generated for each block.
+
+  <subsubsection|Key to extrinsics pairs><label|sect-changes-trie-extrinsics-pairs>
```
drskalman: I won't mention the "pair" anymore, I think people knows hash table or a trie stores a (key, value) pair.I don't think they need to be reminded at every step. I'd just refer to its value such as "list of extrisics"  and drop the word pair from here after.

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+  <\algorithm|<name|Key-To-Block-Pairs>(<math|B<rsub|i>>, interval, levels)>
+    <strong|Initialize:>
+
+    1. <math|I=interval>
+
+    2. <math|L=levels>
+
+    <strong|for each> <math|l\<in\><around*|{|1,\<ldots\>,L|}>>
+
+    3.<space|1em>if <math|H<rsub|i><around*|(|B<rsub|i>|)>=I<rsup|l>>
```
drskalman: use \STATE instead of putting the line numbers manually

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+  block, formally applied as:
+
+  <\algorithm|<name|Key-To-Block-Pairs>(<math|B<rsub|i>>, interval, levels)>
+    <strong|Initialize:>
+
+    1. <math|I=interval>
+
+    2. <math|L=levels>
+
+    <strong|for each> <math|l\<in\><around*|{|1,\<ldots\>,L|}>>
```
drskalman: Use \FOR instead

polkadot-host-spec/c03-transition.tm
drskalman: use \IF instead

polkadot-host-spec/c03-transition.tm
```diff
@@ -670,112 +671,305 @@
...
+    <text|<math|<text|<strong|4<rsup|3>>>>>-nth block, where the pair value
+    contains the block numbers of every block that changed the specified
+    storage key. this level only considers block number of the last 64
+    (<math|=4<rsup|3>>) blocks.
+
+    <\itemize-dot>
+      <item>Example: this level occurs at block 64, 128, 196, 256, 320, etc.
+    </itemize-dot>
+  </enumerate-numeric>
+
```
drskalman: in my opinion there are too many example here. You might one to drop at least one. But your call.
I would also demonstrate the whole calculation for at least one of the example the exact value for the key (type, ?, Key) and the C_Value/A_value.

polkadot-host-spec/c03-transition.tm
drskalman: also in type 2 as we discussed, the B in H_i(B) is the  last block of the interval I think this need to be clarified.

drskalman (changes requested):
OK this is my review for 03-transition.tm I haven't looked at the change trie host api yet. I also made some changes directly. I'm going to commit those as well.

drskalman:
#129 says it is on hold so I'm dropping the waiting-for-review label.

lamafab:
Merging this for now (https://github.com/w3f/polkadot-spec/issues/129#issuecomment-680972945).

Also, added an disclaimer:

image

PR #154, opened 2020-03-27
cpp-tester: Update to changed storage API

FlorianFranzen:
Fixes build failures due to soramitsu/kagome#278


FlorianFranzen:
The cpp-tester is building against the latest version now, however this shows that 5 / 30 test are failing now.

FlorianFranzen:
These patches will most likely be included in #157.

FlorianFranzen:
Included in #179.

Issue #46, opened 2019-09-09
Update Babe Specification in Consensus Chapter


drskalman:
### Resources
 - [Current Research Write Up](https://w3f-research.readthedocs.io/en/latest/polkadot/BABE/Babe.html)

### To be done
 - [x] Specify `NextEpochDescriptor` (#173)
 - [x] Update relative time algorithm (#168)
 - [x] Specify `NextConfigDescriptor` (paritytech/substrate#5514)
 - [x] Review #50, #54, #55, #56

### Beyond current scope
 - ~Specify `SecondarySlot` (#233)~
 - ~Specify `Lateness` (paritytech/substrate#5519)~
 - ~Check Equivocations change (paritytech/substrate#6362)~

rphmeier:
A small write-up here to assist a bit (I had offline conversations with Al & Jeff about this in Zug)

1. The "configuration" of the chain is drawn by querying `BabeApi::configuration` with the state of the genesis block. The result of this function at any other block is unimportant for now (but may become more important as we move into designing things like swappable consensus). The struct is `babe_primitives::BabeConfiguration` - that should be a minimal set of things to get the chain started.
2. At the first block of every epoch, we're expecting a `Consensus` digest item containing the authorities, weights, and randomness for the next epoch. The definition of this can also be found in the primitives crate as `NextEpochDescriptor`.
3. The first epoch (epoch #0) is special-cased. The genesis block is said to have slot 0. The slot of the first block defines the actual slot where epoch 0 starts. Thus any block with number 1 is the first block of the first epoch, and should have a `NextEpochDescriptor` digest to describe epoch 1. This is expected to be the same randomness and authorities as the genesis configuration -- we aren't really worried about byzantine attacks that happen on the first block. The client code should have a mechanism for ensuring that a block with number 1 has a specific hash that is added to the chain-spec once the chain is launched.

rphmeier:
about secondary-slots (i.e. aurababous):
The configuration contains a flag `secondary_slots` (previously, this could change from epoch to epoch, but that doesn't really make sense without the ability to change the slot duration).

If `secondary_slots` is enabled for the chain, then a block at slot `slot_number` can also be sealed round-robin leader based on `i = blake2_256(encode((epoch_randomness, slot_number)) % epoch_authorities.len(); epoch_authorities[i]`, where `epoch_randomness` and `epoch_authorities` are determined based on the parent chain and slot number.

drskalman:
[...]

PR #168, opened 2020-04-14
Update BABE specification (Relative Time, Consensus Message Digest, etc.)

FlorianFranzen:
Updated based on [latest research write up](https://w3f-research.readthedocs.io/en/latest/polkadot/block-production/Babe.html).

### To be done

 - [x] Update algorithm to take probabilistic finalization into account and define `Slot-Time(s)` again
 - [x] Simplify definition of sync epoch (e.g. as multiple of epoch)
 - [x] Review and publish Babe randomness algorithm to runtime spec.
 - [x] Mention that arrival time of only new blocks are valid (in contrast to block received during sync/catch up).
 - [x] Explanation to how to use the algorithm output to estimate future slots.

I assume block arrival times should not be saved to disk, but sync-epochs are rather long (~5h currently) and restarts happen. So for now that would mean it could take validators 5-10h to join the network. This might get even longer if we define sync epochs as a multiple of epochs.

It might also be worth extending the definition of finalized blocks to Grandpa finalized blocks and use the probabilistic finalized blocks only as a fall-back.

This is part of our efforts in #46 and also tracked internally by research [here](https://github.com/w3f/research-internal/issues/110).

Rebased on top of #249 to utilized the improved diffs support.


FlorianFranzen:
TeXmacs Diff

drskalman:
@FlorianFranzen talks to @hndnklnc to clear out the points mentioned so it can be turned into spec.

FlorianFranzen:
@hndnklnc wrote up some changes to the protocol, which are tracked in w3f/research-internal#375 and w3f/research-internal#376. I will update the spec accordingly in the coming days.

burdges:
Added some info for devs in https://github.com/paritytech/substrate/issues/5978 and asked for cumulus to pass clock to parachains in https://github.com/paritytech/cumulus/issues/72#issuecomment-627149633

burdges:
Is relative time working yet?

hndnklnc:
> Is relative time working yet?

not yet

FlorianFranzen:
@burdges @hndnklnc I am getting an issue together with outstanding points to discuss to get an implementation going. Will add you to it once it is ready.

host-spec/c06-consensus.tm
```diff
@@ -24,29 +24,30 @@
...
-  weight <math|W<rsub|A>> in Definition <reference|defn-authority-list>
-  exists for potential improvements in the protocol and could have a use-case
-  in the future.
+  Note that in Polkadot, all authorities have the weight
+  <math|w<rsub|A><rsub|>=1>. The weight <math|w<rsub|A>> in Definition
+  <reference|defn-authority-list> exists for potential improvements in the
+  protocol and could have a use-case in the future.
 
   <subsection|Runtime-to-Consensus Engine
   Message><label|sect-consensus-message-digest>
```
hndnklnc, line 55: Small grammatical issues:
-  The authority lists (see Definition <reference|defn-authority-list>) **is** -> The authority lists (see Definition <reference|defn-authority-list>) **are**

- The runtime inform -> The runtime inform**s**

host-spec/c06-consensus.tm
```diff
@@ -117,12 +150,34 @@
   </definition>
 
   The Polkadot Host should inspect the digest header of each block and
-  delegates consesus messages to their consensus engines. Consensus engine
-  should react based on the type of consensus messages they receives as
-  follows:
+  delegates consesus messages to their consensus engines.
```
hndnklnc: delegates -> delegate

host-spec/c06-consensus.tm
```diff
@@ -117,12 +150,34 @@
...
+
+  The BABE consensus engine should react based on the type of consensus
+  messages they receives as follows:
```
hndnklnc: they -> it

host-spec/c06-consensus.tm
```diff
@@ -220,11 +275,11 @@
     <math|\<cal-E\><rsub|n>> by <math|sc<rsub|n>>. <math|sc<rsub|n>> is set
     to the <verbatim|duration> field in the returned data from the call of
     the Runtime entry <verbatim|BabeApi_configuration> (see
-    <reference|sect-rte-babeapi-epoch>) at the beginning of each epoch. For a
-    given block <math|B>, we use the notation <strong|<math|s<rsub|B>>> to
-    refer to the slot during which <math|B> has been produced. Conversely,
-    for slot <math|s>, <math|\<cal-B\><rsub|s>> is the set of Blocks
-    generated at slot <math|s>.
+    <reference|sect-rte-babeapi-epoch>) at genesis. For a given block
```
hndnklnc, line 289: Why is the parameter `sc` obtained from the genesis? Instead of obtaining  `sc` from the genesis block, obtaining it from e.g. the first block of an epoch makes the parameter update more flexible. For example, if we change the parameter `c` later, we have to change `sc` as well.

host-spec/c06-consensus.tm
```diff
@@ -220,11 +275,11 @@
...
+    <math|B>, we use the notation <strong|<math|s<rsub|B>>> to refer to the
+    slot during which <math|B> has been produced. Conversely, for slot
+    <math|s>, <math|\<cal-B\><rsub|s>> is the set of Blocks generated at slot
+    <math|s>.
   </notation>
 
   Definition <reference|defn-epoch-subchain> provides an iterator over the
```
hndnklnc, line 296: an specific epoch -> a specific epoch

host-spec/c06-consensus.tm
```diff
@@ -241,6 +296,20 @@
 
   <subsection|Block Production Lottery>
 
+  <\definition>
+    <label|defn-babe-constant>The <with|font-series|bold|BABE constant>
+    <math|<with|font-series|medium|c>\<in\><around*|(|0,1<rsub|>|)>> is the
+    probability that a slot will be not be empty. It is initialized at
```
hndnklnc: is the probability that a slot will be not be empty -> is the probability that a slot will not be empty

host-spec/c06-consensus.tm
```diff
@@ -255,10 +324,8 @@
 
     where <math|AuthorityDirectory<rsup|\<cal-E\><rsub|n>>> is the set of
```
hndnklnc, line 338: how does a validator in an epoch obtain <math|AuthorityDirectory<rsup|\<cal-E\><rsub|n>>>? It is important to obtain this set from at latest  epoch E_{n-3}

host-spec/c06-consensus.tm
```diff
@@ -302,28 +369,47 @@
...
 
   <subsection|Slot Number Calculation>
 
-  It is essential for a block producer to calculate and validate the slot
-  number at a certain point in time. Slots are dividing the time continuum in
-  an overlapping interval. At a given time, the block producer should be able
-  to determine the set of slots which can be associated to a valid block
-  generated at that time. We formalize the notion of validity in the
-  following definitions:
+  It is imperative for the security of the network that each block producer
```
hndnklnc, line 400: This sentence is too long and hard to understand. Also, the issue is not to determine the set of slots. It is determining the slot number at a given time. It will be better to rephrase it.

host-spec/c06-consensus.tm
FlorianFranzen: I choose this for the same reason epoch length is fixed at genesis. Substrate currently can not easily handle interval changes. It would be worth while discussing this with the implementers.

host-spec/c06-consensus.tm
```diff
@@ -302,28 +369,47 @@
...
+  clock source (e.g. through the <with|font-shape|italic|Network Time
+  Protocol> or the <with|font-shape|italic|Global Positioning System>). To
+  stay in synchronization each producer is therefore required to periodically
+  estimate its local clock offset in relation to the rest of the network.\ 
+
+  This estimation depends on the two fixed parameters
+  <with|font-series|bold|<math|k>> (def. <reference|defn-prunned-best>) and
+  <math|<with|font-series|bold|s<rsub|cq>>> (def.
+  <reference|defn-chain-quality>). These are choosen based on the results of
+  simulation parameterized to meet certain security bounds, currently
```
hndnklnc: These are choosen based on the results of simulation parameterized-> These are chosen based on the results of formal security analysis

host-spec/c06-consensus.tm
```diff
@@ -302,28 +369,47 @@
...
+  assuming a <math|1 s \ >clock drift per day and targeting a probability
```
hndnklnc: and c= 0.38

host-spec/c06-consensus.tm
```diff
@@ -302,28 +369,47 @@
...
+  resistance against network delay up to <math|<frac*|1|3>> of the slot time.
 
   <\definition>
-    <label|slot-time-cal-tail>The <strong|slot tail>, formally referred to by
-    <math|SlTl> represents the number of on-chain blocks that are used to
-    estimate the slot time of a given slot. This number is set to be 1200.
+    <label|defn-prunned-best>The <with|font-series|bold|prunned best chain>
+    <math|C<rsup|\<#250C\>k>> is the best chain selected according to the
+    best chain selection rule with the last k Blocks prunned. Based on
+    simulation results, we choose <math|k=140>. The
```
hndnklnc: simulation result -> security analysis

host-spec/c06-consensus.tm
```diff
@@ -302,28 +369,47 @@
...
+    <with|font-series|bold|last (probabilistically) finalized block>
+    describes the last block in this prunned best chain.
   </definition>
 
-  Algorithm <reference|algo-slot-time> determines the slot time for a future
-  slot based on the <em|block arrival time> associated with blocks in the
-  slot tail defined in Definition <reference|defn-block-time>.
+  <\definition>
+    <label|defn-chain-quality>The <with|font-series|bold|chain quality>
+    <math|s<rsub|cq>> represents the number of slots hat are used to estimate
```
hndnklnc: hat -> that

host-spec/c06-consensus.tm
```diff
@@ -302,28 +369,47 @@
...
+    the local clock offset. Based on simulation results, we choose
```
hndnklnc: simulation results -> security analysis

host-spec/c06-consensus.tm
```diff
@@ -337,6 +423,25 @@
     there is no ambiguity about the underlying node.
   </definition>
 
+  At the end of each sync-epoch (def. <reference|defn-sync-epoch>) these
+  block arrivel times are used to update the current clock offset using the
```
hndnklnc: arrivel -> arrival

host-spec/c06-consensus.tm
```diff
@@ -337,6 +423,25 @@
...
+  median algorithm (alg. <reference|algo-slot-time>).
+
+  <\definition>
+    <label|defn-sync-epoch>A <with|font-series|bold|sync epoch> (unrelated to
```
hndnklnc: It would be nice to give an algorithm to determine the start of the sync epoch: e.g., store the last slot number s_i-1 in the previous epoch, check the difference between the slot number of last finalized block (s_i) and s_i-1 when BT is updated. If the difference is greater than or equal to s_{cq}, run output the  s_i

host-spec/c06-consensus.tm
```diff
@@ -337,6 +423,25 @@
...
+    the epoch defined in <reference|defn-epoch-slot>) is the inteval at which
```
hndnklnc: inteval -> interval

host-spec/c06-consensus.tm
```diff
@@ -337,6 +423,25 @@
...
+    that <math|<wide|s|\<bar\>><rsub|i><rsub|>-<wide|s|\<bar\>><rsub|i-1>\<geq\>s<rsub|cq>>
+    where <math|<wide|s|\<bar\>><rsub|i-1>> is the slot number of the last
+    (probabilistically) finalized block in the previous sync-epoch
+    <math|\<frak-E\><rsub|i-1>>. If the previous epoch is the first epoch
+    <math|\<frak-E\><rsub|1>> then <math|<wide|s|\<bar\>><rsub|i-1>=<wide|s|\<bar\>><rsub|i>=0>.
+  </definition>
+
   In addition to the arrival time of block <math|B>, the block producer also
   needs to know how many slots have passed since the arrival of <math|B>.
   This value is formalized in Definition <reference|defn-slot-offset>.
```
hndnklnc: Why don't we give a formula for the slot-offset? I guess it should be If k < l, (l - k- 1) * eplen + eplen - i + j, if k = l, j - i

host-spec/c06-consensus.tm
```diff
@@ -302,28 +369,47 @@
...
+    <math|s<rsub|cq>> represents the number of slots hat are used to estimate
+    the local clock offset. Based on simulation results, we choose
+    <math|s<rsub|cq>=3000>.
+  </definition>
+
+  The prerequisite for such a calculation is that each producer stores the
+  arrival time of each block (def. <reference|defn-block-time>).\ 
 
   <\definition>
     <label|defn-block-time>The <strong|block arrival time> of block <math|B>
```
hndnklnc, line 475: Maybe, we should emphasize that the local timer is the computer clock which is not adjusted by any protocol such as NTP

host-spec/c06-consensus.tm
```diff
@@ -351,25 +456,20 @@
...
       </state>
 
       <\state>
-        <math|B<rsub|d>\<leftarrow\>><name|Deepest-Leaf(<math|BT>)>
-      </state>
-
-      <\state>
-        <FOR-IN|<math|B<rsub|i>>|<name|SubChain>(<math|B<rsub|H<rsub|n><around*|(|B<rsub|d>|)>-SITL>>,
-        <math|B<rsub|d>>)>
+        <FOR-IN|<math|B<rsub|i>>|<math|\<frak-E\><rsub|j>><name|>>
```
hndnklnc, line 521: the blocks of the epoch are not defined anywhere.  It should be all blocks that are probabilistically finalized during the sync-epoch

host-spec/c06-consensus.tm
```diff
@@ -351,25 +456,20 @@
   </definition>
 
   <\algorithm>
-    <label|algo-slot-time><name|Slot-Time>(<math|s>: the slot number of the
-    slot whose time needs to be determined)
+    <label|algo-slot-time><name|Median-Algorithm>(<math|\<frak-E\><rsub|j>>:
```
hndnklnc, line 511: We should say that the median algorithm has to be run at the beginning of a sync epoch

host-spec/c06-consensus.tm
```diff
@@ -420,7 +525,7 @@
 
     \;
 
-    <math|H<rsub|BABE><around*|(|B|)>> must be included as a diegst item of
+    <math|H<rsub|BABE><around*|(|B|)>> must be included as a digest item of
     Pre-Runtime type in the header digest <math|H<rsub|d><around*|(|B|)>> as
     defined in Definition <reference|defn-digest>.\ 
   </definition>
```
hndnklnc, line 584: what is H_h(B)?

hndnklnc: In Algorithm 6.3., SLOT_TIME(s) is not defined and threshold \tau should be written as defined in Definition 6.8

host-spec/c06-consensus.tm
```diff
@@ -505,41 +610,35 @@
...
 
-  At the end of epoch <math|\<cal-E\><rsub|n>>, each block producer is able
-  to compute the randomness seed it needs in order to participate in the
-  block production lottery in epoch <math|\<cal-E\><rsub|n+2>>. For epoch 0
-  and 1, the randomness seed is provided in the genesis state. The
-  computation of the seed is described in Algorithm
-  <reference|algo-epoch-randomness> which uses the concept of epoch subchain
-  described in Definition <reference|defn-epoch-subchain>.
+  In the beginning of each epoch <math|\<cal-E\><rsub|n>> the host will
+  receives the randomness seed <math|\<cal-R\><rsub|\<cal-E\><rsub|n+1>>>(def.
```
hndnklnc: receives -> receive

host-spec/c06-consensus.tm
```diff
@@ -505,41 +610,35 @@
...
   <subsection|Epoch Randomness><label|sect-epoch-randomness>
 
-  At the end of epoch <math|\<cal-E\><rsub|n>>, each block producer is able
-  to compute the randomness seed it needs in order to participate in the
-  block production lottery in epoch <math|\<cal-E\><rsub|n+2>>. For epoch 0
-  and 1, the randomness seed is provided in the genesis state. The
-  computation of the seed is described in Algorithm
-  <reference|algo-epoch-randomness> which uses the concept of epoch subchain
-  described in Definition <reference|defn-epoch-subchain>.
+  In the beginning of each epoch <math|\<cal-E\><rsub|n>> the host will
```
hndnklnc, line 675: This paragraph is not clear and it seems it is not correct. Why does the host have to receive R_E_{n+1} in the beginning of E_n to participate E_n+1. In the beginning of E_n+1, the host has to obtain the randomness using the VRF outputs belonging to the probabilistically finalized blocks of E_{n-1}.  The previous version was correct.

host-spec/c06-consensus.tm
```diff
@@ -505,41 +610,35 @@
...
 
-      <\state>
-        <FOR-IN|<math|B>|><em|<name|SubChain(<math|\<cal-E\><rsub|n-2>>)>>
-      </state>
+  \;
 
-      <\state>
-        <math|\<rho\>\<leftarrow\>\<rho\><around*|\|||\|>d<rsub|B>><END>
-      </state>
+  <\definition>
```
hndnklnc, line 667: I did not understand here.

host-spec/c06-consensus.tm
FlorianFranzen: As these are not (any longer?) computed by the host, but by the runtime, I moved any of those details to the runtime spec. I will leave this comment till it is under review.

host-spec/c06-consensus.tm
FlorianFranzen: Either that or simplifying the definition of a sync-epoch so far that this no longer becomes necessary.

host-spec/c06-consensus.tm
FlorianFranzen: Yes, this is another reason to simplify this definition.

host-spec/c06-consensus.tm
FlorianFranzen: As this spec is only concerned with the host side of things and both the authority set and epoch randomness are given to the host in the same consensus message, this description should be correct.

host-spec/c06-consensus.tm
FlorianFranzen: Sentence was completely removed from definition after discussion with Syed.

host-spec/c06-consensus.tm
FlorianFranzen: Good catch. I think this sentence is still from the time where I assumed the timestamp in a block are actually checked by every validator. Will update accordingly.

host-spec/c06-consensus.tm
FlorianFranzen: Added "... measured by a clock that is otherwise not adjusted by any external protocol."

host-spec/c06-consensus.tm
FlorianFranzen: > what is H_h(B)?

It is the block header hash, but I also added a note to the definition of the block signature to  make this clearer.

FlorianFranzen: > In Algorithm 6.3., SLOT_TIME(s) is not defined

Indeed I seem overlooked that when changing the name of the median algorithm definition in 6.2. I will rework the definition to have that fixed.

> threshold \tau should be written as defined in Definition 6.8

Do you mean adding the subscript or replacing \tau with its definition?

host-spec/c06-consensus.tm
FlorianFranzen: Indeed, the algorithm just ignores that fact. I will update it accordingly.

host-spec/c06-consensus.tm
```diff
@@ -24,41 +24,42 @@
...
+    Definition <reference|defn-session-key>. And <math|w<rsub|A>> is a
+    <verbatim|u64> value, indicating the authority weight. The value of
+    <math|Auth<rsub|C><around*|(|B|)>> is part of the Polkadot state. The
+    value for <math|Auth<rsub|C><around*|(|B<rsub|0>|)>> is set in the
+    genesis state (see Section <reference|sect-genesis-block>) and can be
+    retrieved using a runtime entery corresponding to consensus engine
+    <math|C>.
   </definition>
 
-  Note that in Polkadot, all authorities have the weight equal to 1. The
```
drskalman, line 46: I would make this more prominent. This makes the implementation significantly less complicated and we don't want anybody to miss this out I think you can use \note or \remark.

host-spec/c06-consensus.tm
```diff
@@ -241,6 +297,20 @@
 
   <subsection|Block Production Lottery>
 
+  <\definition>
+    <label|defn-babe-constant>The <with|font-series|bold|BABE constant>
+    <math|<with|font-series|medium|c>\<in\><around*|(|0,1<rsub|>|)>> is the
+    probability that a slot will not be empty. It is initialized at genesis
```
drskalman: This the motivation behind the definition I'd take it out of the definition and describe it as an after remark. If I'm an implementer and read this my first reaction is that I have to somehow do some probablity calculation related to this (which is not the case) so let keep the definition simpler.

host-spec/c06-consensus.tm
```diff
@@ -241,6 +297,20 @@
...
+    using the value returned by a call <verbatim|BabeApi_configuration>(see
+    <reference|sect-rte-babeapi-epoch>) and then can be updated by the
```
drskalman: I had a conversation about this. BabeApi_configuration actually always returns the correct updated value, it is just easier and more efficient to use the message digest (the same as for the authority set). so this need to be clarified. Plus I'd debate if it is a good idea to include all the detail of how to retrieve the value in the definition. Personally I'd keep the definiton simpler and add the extra information after.

host-spec/c06-consensus.tm
```diff
@@ -241,6 +297,20 @@
...
+    runtime for the next epoch through the \PNext Config Data\Q consensus
+    message digest (see def. <reference|defn-consensus-message-digest>) in
+    the first block of each epoch. It is encoded as a tuple of two unsigned
+    64 bit integers <math|<around*|(|c<rsub|nominator>,c<rsub|denominator>|)>>
+    which are used to compute the rational
```
drskalman: This really should go into definition of the particular runtime-consensu message (the encoding shouldn't be part of definition of c, if you change the encoding of c it will still remains as c).

drskalman:
Let ask @hndnklnc about the "Mention that arrival time of only new blocks are valid (in contrast to block received during sync/catch up)." and how a node can distinguish between catchup or actual new block.

hndnklnc:
> Let ask @hndnklnc about the "Mention that arrival time of only new blocks are valid (in contrast to block received during sync/catch up)." and how a node can distinguish between catchup or actual new block.

What is catchup block?

drskalman (commented):
> What is catchup block?

:-) The problem is exactly that. How to define from BABE prespective that if the block is freshly out of own or it has been baked a while ago.
I'm not sure if there is a formal definition, but if you go offline and then you receives blocks that have been produced in the past. The problem is that I'm not sure how you know if what you get is a 'fresh' block or you are catching up to what you missed (in the sense that you were not online when first this block has been propagated by your peers). There is a formal definition in grandpa but that doesn't apply to BABE.

burdges:
At least validators could define a catchup block using the CRDT inside the slashing reform certificate wrapper, so its nolonger a catch up block after the chain "sufficiently finalizes" your fresh certificate wrapper.  See https://hackmd.io/FY4MCfZCSZGkV-Ulo3VSRg

We've only thought closely about GRANDPA, where "sufficiently finalized" means two or three blocks after finality.  I think here merely appearing on-chain suffices, so no finality at all, which meshes better with BABE anyways.  A fresh certificate wrapper does not change the BABE key of course, altering a BABE key requires waiting one epoch.

track https://github.com/paritytech/substrate/issues/7398  
cc https://github.com/w3f/research-internal/issues/530

hndnklnc:
> > What is catchup block?
> 
> :-) The problem is exactly that. How to define from BABE prespective that if the block is freshly out of own or it has been baked a while ago.
> I'm not sure if there is a formal definition, but if you go offline and then you receives blocks that have been produced in the past. The problem is that I'm not sure how you know if what you get is a 'fresh' block or you are catching up to what you missed (in the sense that you were not online when first this block has been propagated by your peers). There is a formal definition in grandpa but that doesn't apply to BABE.

There is not any mechanism in BABE to catchup blocks if you were online at some point. I think GRANDPA handles it. I assume you want to understand what happens in the relative time protocol in this case. If a validator was offline, normally it should start the relative time protocol in the next sync-epoch just after getting all catch-up blocks. However, since waiting for the sync-epoch is waste of time (@FlorianFranzen made me realize this problem), I suggested [this solution](https://cseweb.ucsd.edu/~mihir/papers/multisignatures-ccs.pdf) for temporarily clock adjustment which is not perfect. So, we have another issue to fix some problems in the temporary clock adjustment mechanism https://github.com/w3f/polkadot-spec/issues/263#issue-680261122

host-spec/c06-consensus.tm
drskalman: Hmm don't we have a section on how to obtain the AuthorityDirectory? It should be obtained from the runtime because it is determined by the chain state (elections and slashing). Nonetheless runtime send message to host by means of digest to tell the host about the changes in the authset.

host-spec/c06-consensus.tm
```diff
@@ -302,28 +370,49 @@
...
+
+  This estimation depends on the two fixed parameters
+  <with|font-series|bold|<math|k>> (def. <reference|defn-prunned-best>) and
+  <math|<with|font-series|bold|s<rsub|cq>>> (def.
+  <reference|defn-chain-quality>). These are choosen based on the results of
+  formal security analysis, currently assuming a <math|1 s \ >clock drift per
+  day and targeting a probability lower than <math|0.5%> for an adversary to
+  break BABE in 3 years with a resistance against network delay up to
+  <math|<frac*|1|3>> of the slot time and a Babe constant (def.
+  <reference|defn-babe-constant>) of <math|c=0.38>.
```
drskalman: We do not offer any security claim or details in the spec. It suffices here to say that they are chosen to the security reason and cite the research. I don't think security argument is part of the spec.

host-spec/c06-consensus.tm
```diff
@@ -302,28 +370,49 @@
...
 
   <\definition>
-    <label|slot-time-cal-tail>The <strong|slot tail>, formally referred to by
-    <math|SlTl> represents the number of on-chain blocks that are used to
-    estimate the slot time of a given slot. This number is set to be 1200.
+    <label|defn-prunned-best>The <with|font-series|bold|prunned best chain>
+    <math|<with|font-series|bold|C<rsup|\<#250C\>k>>> is the best chain
+    selected according to the best chain selection rule with the last k
```
drskalman: There is a formal definition of best chain in the spec, you could refer to it instead of refering to best chain selection rule.

host-spec/c06-consensus.tm
```diff
@@ -337,6 +426,25 @@
     there is no ambiguity about the underlying node.
   </definition>
 
+  At the end of each sync-epoch (def. <reference|defn-sync-epoch>) these
+  block arrival times are used to update the current clock offset using the
+  median algorithm (alg. <reference|algo-slot-time>).
+
+  <\definition>
+    <label|defn-sync-epoch>A <with|font-series|bold|sync epoch> (unrelated to
```
drskalman: We deсided not to have sync epoch. and use the epoch period for the same function for the sake of simplifying the protocol.

host-spec/c06-consensus.tm
```diff
@@ -420,14 +528,15 @@
...
+    <math|H<rsub|BABE><around*|(|B|)>> must be included as a digest item of
     Pre-Runtime type in the header digest <math|H<rsub|d><around*|(|B|)>> as
     defined in Definition <reference|defn-digest>.\ 
   </definition>
 
   <\definition>
     <label|defn-block-signature><label|defn-babe-seal>The <strong|Block
-    Signature> noted by <math|S<rsub|B>> is computed as
+    Signature> <math|S<rsub|B>> is a signature of the block header hash (see
+    def.<reference|defn-block-header-hash>) and defined as
```
drskalman: We have been using (see Definition ...) instead of (see def.) and perhaps we should stick to that convention for the sake of consistency.

host-spec/c06-consensus.tm
```diff
@@ -505,43 +614,21 @@
...
+    <label|defn-epoch-randomness>For epoch <math|\<cal-E\>> there is a
+    32-byte <with|font-series|bold|randomness seed>
+    <math|\<cal-R\><rsub|\<cal-E\>>> computed based on the previous epochs
+    VRF outputs. For <math|\<cal-E\><rsub|0>> and <math|\<cal-E\><rsub|1>>,
+    the randomness seed is provided in the genesis state.
+  </definition>
 
-  \;
+  In the beginning of each epoch <math|\<cal-E\><rsub|n>> the host will
+  receive the randomness seed <math|\<cal-R\><rsub|\<cal-E\><rsub|n+1>>>(def.
```
drskalman: again def. -> Definition

host-spec/c06-consensus.tm
drskalman: I'm curious to know how the runtime is computing the randomness, are the VRF outputs are stored in the state storage or whole blocks are stored there, could you point me to the code in the runtime for this computation?

drskalman (dismissed):
Still we need to come up with some stale vs fresh, catch up vs live production notion for BABE.

drskalman:
> There is not any mechanism in BABE to catchup blocks if you were online at some point. I think GRANDPA handles it. I assume you want to understand what happens in the relative time protocol in this case. If a validator was offline, normally it should start the relative time protocol in the next sync-epoch just after getting all catch-up blocks. However, since waiting for the sync-epoch is waste of time (@FlorianFranzen made me realize this problem), I suggested [this solution](https://cseweb.ucsd.edu/~mihir/papers/multisignatures-ccs.pdf) for temporarily clock adjustment which is not perfect. So, we have another issue to fix some problems in the temporary clock adjustment mechanism [#263 (comment)](https://github.com/w3f/polkadot-spec/issues/263#issue-680261122)

@hndnklnc  What do you mean by "after getting all catch-up blocks" how do I know the blocks I'm getting are catch up blocks and when the new fresh blocks are coming? You tell your peer last block I've seen is n, your peers says I have block till n+m and then send you blocks n+1 to n+m this happens all the time, I believe. The only difference is that maybe at some point you gets the block 1 at the time. Should that be the definition of when you should trust the arrival time?

hndnklnc:
> > There is not any mechanism in BABE to catchup blocks if you were online at some point. I think GRANDPA handles it. I assume you want to understand what happens in the relative time protocol in this case. If a validator was offline, normally it should start the relative time protocol in the next sync-epoch just after getting all catch-up blocks. However, since waiting for the sync-epoch is waste of time (@FlorianFranzen made me realize this problem), I suggested [this solution](https://cseweb.ucsd.edu/~mihir/papers/multisignatures-ccs.pdf) for temporarily clock adjustment which is not perfect. So, we have another issue to fix some problems in the temporary clock adjustment mechanism [#263 (comment)](https://github.com/w3f/polkadot-spec/issues/263#issue-680261122)
> 
> @hndnklnc What do you mean by "after getting all catch-up blocks" how do I know the blocks I'm getting are catch up blocks and when the new fresh blocks are coming? You tell your peer last block I've seen is n, your peers says I have block till n+m and then send you blocks n+1 to n+m this happens all the time, I believe. The only difference is that maybe at some point you gets the block 1 at the time. Should that be the definition of when you should trust the arrival time?

I was always assuming that there is a mechanism (with GRADPA) for the new joining validators to make the validator learn the existing blocks (catchup blocks) to be updated and validators actively join block production after this learning process. If it does not exist, yes you are right we need to come up with an algorithm that differentiates them

lamafab:
@FlorianFranzen works on this partially. No clear estimate for completion.

burdges:
I've added more detail for relative time and approval assignments in https://github.com/paritytech/substrate/issues/7398 but maybe we should extract all this onto one design document, either expanding https://hackmd.io/FY4MCfZCSZGkV-Ulo3VSRg or something new, maybe even being more formal.

host-spec/c06-consensus.tm
FlorianFranzen: Replaced this with a reference to `defn-longest-chain`.

host-spec/c06-consensus.tm
FlorianFranzen: I made this less obtrusive, but I did not remove this, as this is only published in a mutual form, to which I can not really link reliably.

host-spec/c06-consensus.tm
FlorianFranzen: Done, moved it to definition 6.3.

host-spec/c06-consensus.tm
FlorianFranzen: Ok, added some more details about that.

host-spec/c06-consensus.tm
FlorianFranzen: I added another sentence explaining what it is used for and moved out any additional info into remarks.

host-spec/c06-consensus.tm
FlorianFranzen: Yes, this is all part of the update.

host-spec/c06-consensus.tm
FlorianFranzen: Addressed by simplifying the algorithm.

host-spec/c06-consensus.tm
FlorianFranzen: Added this requirement to the text now.

host-spec/c06-consensus.tm
FlorianFranzen: Added.

host-spec/c06-consensus.tm
FlorianFranzen: Added back Slot-Time and updated threshold.

host-spec/c06-consensus.tm
FlorianFranzen: This was moved to the runtime specification, as the randomness is now computed in the runtime.

PR #169, opened 2020-04-16
Spec polite grandpa round limit

### LINKED ISSUE #132: Spec Catchup response message
catch up response is send in reply to catch up request and needs to be specced
as part of grandpa. They are implemented in gossip.rs

### LINKED ISSUE #134: Spec GRANDPA catch up procedure
This include speccing:
- [ ] The condition which merits sending a catchup request.
- [ ] Which a cachup request desreves a response.
- [ ] catchup request message.
- [ ] catchup response message.
- [ ] Algorithm to process a catch up request and generate a response.
- [ ] Algorithm to process a catch up response and how to update the state accordingly.

drskalman:


lamafab:
@drskalman With what tool do I generate the result of the UML?

Catchup message review is here -> https://github.com/w3f/polkadot-spec/issues/132#issuecomment-617881385

PR #175, opened 2020-04-23
Spec importing block procedure

### LINKED ISSUE #174: Spec importing block procedure
This has come out of the comment in the channel:
"
-okay, I found that inconsistency between states of full and syncing nodes is causes by validate_transaction because it changes the state and we don't execute that on syncing node
if we invoked validate_transaction should we use the state before validate_transaction for initialize_block, apply_extrinsics and so on?
another way to put that: when our node is slot leader, we should create a new block on the state that we ended up with after the best imported block, right?

-When do you call validate_transaction? You only need to call this when an external transaction comes it. While block building that is not called explictly
And every time we call validate_transaction, we throw away any state changes

I call it when validating node receives transaction from RPC

- Yeah that is correct :)

- And every time we call validate_transaction, we throw away any state changes
Thanks, this is what I wanted to make sure about

- The point is, only when import a block, the state changes are persisted Every other call into the runtime is just thrown away.

- this really should be in the spec as the general rule. I agree.
"

So all the points above need to be reflected in the spec:
- [x] The only persistant changes to the state is due to import block any other changes should be thrown away. We need to make it clear that is the host duty to revert the state on every runtime call.
- [x] So import block algorithm should have store state.
- [x] Validate transaction should specifically say to undo the state. (this probably happens for finalize block or not, do we import our own block explicitly?) At least if we don't finalize block we need to manually revert the state..
- [x] The runtime section should say any changes to state should be wiped out after the call except for import block.

lamafab:
Fixes #174. Clarifies when state changes are thrown away or persisted. Also expands the import block algorithm.

### DIFF
--- a/polkadot-host-spec/ag-runtimeentries.tm
+++ b/polkadot-host-spec/ag-runtimeentries.tm
@@ -92,18 +92,10 @@
   <assign|figure-text|<macro|Figure>>
 
   The following sections describe the standard based on which the Polkadot
-  Host communicates with each runtime entry.
-
-  <section|Managing state changes>
-
-  Only the two runtime functions <verbatim|Core_execute_block> as described
-  in secion <reference|defn-rt-core-execute-block> and
-  <verbatim|BlockBuilder_finalize_block> as described in section
-  <reference|defn-rt-blockbuilder-finalize-block> keep persistant storage.
-  Any state changes created by directly calling the other functions conducted
-  by the Polkadot Host must be disregarded and thrown away after execution.
-  It is therefore the responsibility of the Polkadot Host to wipe any changes
-  created after the those calls.
+  Host communicates with each runtime entry. Do note that any state changes
+  created by calling any of the Runtime functions are not necessarily
+  persisted. See Section <reference|sect-handling-runtime-state-update> for
+  more information.
 
   <section|Argument Specification>
 
@@ -115,7 +107,7 @@
   <\verbatim>
     \ \ \ \ (func $generic_runtime_entry
 
-    \ \ \ \ \ \ (param $data i32) (parm $len i32) (reslut i64))
+    \ \ \ \ \ \ (param $data i32) (parm $len i32) (result i64))
   </verbatim>
 
   \;
@@ -125,6 +117,8 @@
   can similarly either point to the SCALE encoded data the function returns
   (See Sections <reference|sect-runtime-send-args-to-runtime-enteries> and
   <reference|sect-runtime-return-value>).
+
+  \;
 
   In this section, we describe the function of each of the entries alongside
   with the details of the arguments and the return values for each one of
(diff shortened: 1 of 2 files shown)


drskalman:
Could you also make sure that #178 is addressed? Thanks a lot.

lamafab:
> Could you also make sure that #178 is addressed? Thanks a lot.

Updated.

polkadot-host-spec/ag-runtimeentries.tm
```diff
@@ -94,6 +94,17 @@
   The following sections describe the standard based on which the Polkadot
   Host communicates with each runtime entry.
 
+  <section|Managing state changes>
```
drskalman: The appendix isn't really for explaining general aspect of the runtime entries but is for listing specific enteries and their detail. The implementer should get the general rules for implemeting the host from reading the main body of the spec and just comes to the appendix when it needs to call a specific entry. I didn't see this so I added a sub section in the Runtime subsection of the chapter 3. You could move it to the relevant entries in this appendix or merge it with my explanation in the main body of the spec.

polkadot-host-spec/ag-runtimeentries.tm
```diff
@@ -153,7 +164,10 @@
 
   This function should be called when a fully complete block is available
   that is not actively being built on, such as blocks received from other
-  peers.
+  peers. Any state changes created by this function on successful execution
+  are persisted. Additionally, the seal digest in the block header as
```
drskalman: new paragraph perhapse.

polkadot-host-spec/ag-runtimeentries.tm
```diff
@@ -153,7 +164,10 @@
...
+  described in section <reference|defn-digest> must be removed manually
+  before submitting the block.
```
drskalman, line 169: Similarly, this should be mentioned in Chapter 3 where we define seal digest as well as in  Algorithm 3.4. Import-and-Validate-Block and Algorithm 6.5. Verify-Authorship-Right. (as the ticket suggest).

polkadot-host-spec/c03-transition.tm
```diff
@@ -626,8 +647,12 @@
       </state>
 
       <\state>
-        <math|S\<leftarrow\>> <name|Call-Runtime-Entry><math|<around*|(|<text|<verbatim|Core_execute_block>>,B|)>>
+        <math|R\<leftarrow\>> <name|Call-Runtime-Entry><math|<around*|(|<text|<verbatim|Core_execute_block>>,B|)>>
       </state>
+
+      <space|1em>10:<space|1em>if <em|R> = <name|True>
+
+      <space|1em>11:<space|3em><name|Persist-state>
```
drskalman: Persist-state needs an explanation (maybe at the bottom of the algorithm).

drskalman (changes requested):
Thanks a lot.

polkadot-host-spec/ag-runtimeentries.tm
lamafab: @drskalman The removal of the seal digest before submitting to the Runtime is already mentioned in the definition of the Seal digest. Mentioning it here again  does not hurt, imo (since Soramitsu skipped over it).

I'll expand the algorithms, though.

drskalman (approved):

PR #157, opened 2020-03-31
New testing binary for the Host API overhaul

lamafab:
Thanks to @FlorianFranzen, who fixed the building issues, this can be finally merged. The Rust code bases have been separated between `rust_tester` and `rust_tester_legacy` (which still uses the deprecated API).

**Note**: The Kagome tester returns false values. The team has been notified and should fix this soon. The `kagome_tester` is currently commented-out.

This contains tests for the following, updated functions:
```
"ext_storage_set_version_1",
"ext_storage_get_version_1",
"ext_storage_child_set_version_1",
"ext_storage_child_get_version_1",
"ext_storage_read_version_1",
"ext_storage_child_read_version_1",
"ext_storage_clear_version_1",
"ext_storage_child_clear_version_1",
"ext_storage_child_storage_kill_version_1",
"ext_storage_exists_version_1",
"ext_storage_child_exists_version_1",
"ext_storage_clear_prefix_version_1",
"ext_storage_child_clear_prefix_version_1",
"ext_storage_root_version_1",
"ext_storage_child_root_version_1",
"ext_storage_next_key_version_1",
"ext_storage_child_next_key_version_1",
"ext_crypto_ed25519_public_keys_version_1",
"ext_crypto_ed25519_generate_version_1",
"ext_crypto_ed25519_sign_version_1",
"ext_crypto_ed25519_verify_version_1",
"ext_crypto_sr25519_public_keys_version_1",
"ext_crypto_sr25519_generate_version_1",
"ext_crypto_sr25519_sign_version_1",
"ext_crypto_sr25519_verify_version_1",
"ext_hashing_keccak_256_version_1",
"ext_hashing_sha2_256_version_1",
"ext_hashing_blake2_128_version_1",
"ext_hashing_blake2_256_version_1",
"ext_hashing_twox_256_version_1",
"ext_hashing_twox_128_version_1",
"ext_hashing_twox_64_version_1",
"ext_allocator_malloc_version_1",
"ext_allocator_free_version_1",
"ext_trie_blake2_256_root_version_1",
"ext_trie_blake2_256_ordered_root_version_1",
```

PR #179, opened 2020-04-28
CI Rewrite

### LINKED ISSUE #158: Split and cache CI pipeline
Currently there is only one large job that builds all implementations and testers and then runs the actual test. As a result feedback is slow and failures require you to scroll to walls of texts.

Furthermore, the repository mixes documents and code, which led to the solution to only run the full CI on code changes. This creates a lot of false positive CI results and the detection is unreliable (e.g.  does not run when implementation change, only looks at the last commit etc.)

I would therefor like to discuss the following actions:

- [x] Split code and documentation again*
- [x] Add CI job to check and build documents
- [x] Split CI job into building and testing*
- [x] Split testing job by implementation* and test type
- [x] Cache build results with reliable change detection*
- [x] Stop the use of submodules (as we already do for go-tester)*

I slightly adapted the Nix build files I am using to run the tests on my machine into a proof of concept on our [gitlab instance](https://gitlab.w3f.tech/florian/spectest). All the point marked with a __*__ above are already implemented there. This includes a switch to [gitlab's builtin CI](https://gitlab.w3f.tech/florian/spectest/pipelines/1327).
Building is done in nix with the help of nixpkgs. Build results are cached on cachix.

Some great resources to get started with nix is [nixcloud's interactive tour](https://nixcloud.io/tour) and the [Learn NixOS page](https://nixos.org/learn.html) which links to the manuals and other great resources like the Nix Pills or Cookbook.

FlorianFranzen:
This pull request address and fixes #158 and includes #101, #154 and #157.

Most notable changes:
- Move from Circle CI to Github Actions
- Split building and testing split into separate jobs
- Split building by implementation 
- Split building and testing by implementation and fixture
- Add unified Julia script `runtests.jl` to run selected fixtures
- Add unified fixture structure and glue code
- Add cache for build artifact and dependencies
- Add CI pipeline that builds documentation 
- Removes git submodules
- Renames tester to adapter
- Update all adapter to latest implementation release
- Include wasm binary inside substrate-adapter
- Split legacy adapter to separate cargo crate
- Merge and fix kagome hostapi-legacy adaption
- Various small tweaks and fixes

I think we should also start tagging spec releases and use the CI to build pdf diffs that are push to releases, but I will leave that for now. I did not have the time to figure out how to call TexMacs diff function from the command line.


FlorianFranzen:
The kagome adapter still [does not build on Ubuntu](https://github.com/w3f/polkadot-spec/runs/627269082):
```
 /usr/bin/g++-8  -DSPDLOG_COMPILED_LIB -DXXH_EXPORT -I/home/runner/.hunter/_Base/a5dcae9/354b63e/5ef096c/Build/kagome/Install/include/kagome -isystem /home/runner/.hunter/_Base/a5dcae9/354b63e/5ef096c/Install/include -isystem /home/runner/.hunter/_Base/a5dcae9/354b63e/5ef096c/Install/include/binaryen  -O3 -DNDEBUG   -Ded25519_EXPORTS -std=c++17 -o CMakeFiles/kagome-adapter.dir/src/scale_codec.cpp.o -c /home/runner/work/polkadot-spec/polkadot-spec/test/adapters/kagome/src/scale_codec.cpp
In file included from /home/runner/work/polkadot-spec/polkadot-spec/test/adapters/kagome/src/scale_codec.cpp:16:
/home/runner/.hunter/_Base/a5dcae9/354b63e/5ef096c/Install/include/kagome/common/hexutil.hpp:14:10: fatal error: outcome/outcome.hpp: No such file or directory
 #include "outcome/outcome.hpp"
```
I am pretty sure the included header is located in `/home/runner/.hunter/_Base/a5dcae9/354b63e/5ef096c/Build/kagome/Install/include/kagome`, as hunter install it there itself:
```
-- Installing: /home/runner/.hunter/_Base/a5dcae9/354b63e/5ef096c/Build/kagome/Install/include/kagome/outcome/outcome.hpp
```
What, apart from a broken compiler, could be the cause for this error? The `-I` flag seems to be just ignored and it builds [just fine under nix](https://github.com/w3f/polkadot-spec/runs/627269137). 

The [same error](https://github.com/w3f/polkadot-spec/runs/629238943) happens with clang. I can only assume `-isystem` includes are not allow to use `-I` includes and this is actually caused by kagome target not returning the kagome subfolder under `INTERFACE_INCLUDE_DIRS`.

My latest theory is that it is related to this:
> If a standard system include directory, or a directory specified with -isystem, is also specified with -I, the -I option is ignored. The directory is still searched but as a system directory at its normal position in the system include chain. This is to ensure that GCC’s procedure to fix buggy system headers and the ordering for the #include_next directive are not inadvertently changed. If you really need to change the search order for system directories, use the -nostdinc and/or -isystem options.

lamafab (approved):
LGTM. Certain concerns were fixed/clarified directly.

test/README.md
```diff
@@ -1,30 +1,64 @@
...
+- _julia_ to run the testsuite
+- _rust-nightly_ (with wasm target) to build substrate 
+- _cmake_ to build kagome
+- _go_ to build gossamer
+
+While the official target of our testsuite are currently only debian-based systems, there is in general no reason for it to not be able to run on  any recent GNU/Linux or even UNIX-based OS, like OS X.
+
+## General Build
+
+Each of the API adapters has to be build (see [adapters subfolder](./adapters/)), before the testsuite can be run.
```
drskalman: I'm not sure what is the advantage of not havig the option to build and run all tests locally.

build_test.sh
```diff
@@ -1,14 +0,0 @@
-#!/bin/bash
```
drskalman, line 1: I think it was very useful that one could run all the tests locally without dependening on CI infrastructure. It would be good to have build_test.sh back which builds everything locally.

CMakeLists.txt
```diff
@@ -1,31 +0,0 @@
-cmake_minimum_required (VERSION 3.0)
```
drskalman, line 1: if you prefer to have a home brew Makefile for making the whole test suite like what runtime verification has that's OK. But I think it is important to have build system for the whole suit.

drskalman (changes requested):
Thanks a lot. It is a huge refactoring and is much cleaner now. I think we need a Makefile which can replace the old CMake file so one can easily test everything locally. Unless there is an easy way of running the CI locally (which can perhapse be wrapped in a Makefile or a script).

test/adapters/substrate/adapter/wasm-adapter/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
drskalman: this is out of its place as it has nothing to do with substrate. All hosts should support wasm. So maybe we need test/testers or test/test_helpers to keep the universal bit of testing machine.

test/fixtures/AdapterTests.jl
```diff
@@ -0,0 +1,142 @@
+module AdapterTests
```
drskalman, line 1: This is not a test fixture so it should go to some other folder maybe test-helpers testers julia-tester etc.

test/fixtures/HostAPITests.jl
```diff
@@ -0,0 +1,152 @@
+module HostAPITests
```
drskalman, line 1: Same for this.

drskalman (changes requested):
some folder re-arrangement to for the testers' structure to stay logically consistent.

build_test.sh
FlorianFranzen: Ok, I added a Makefile to do just that. Now you can build and run the whole test suite with `make test`.

test/adapters/substrate/adapter/wasm-adapter/Cargo.toml
FlorianFranzen: I agree that the name `wasm-adapter` is confusing. I renamed it to `substrate-adapter-runtime`.  After all it is adapter specific, as only the substrate adapter does not call the Host API directly but through a thin wasm wrapper.

CMakeLists.txt
FlorianFranzen: As above, I added a Makefile for that.

test/README.md
FlorianFranzen: You always had that option, but I agree it was a bit convoluted having to go to each of the adapter subfolders to build the adapters. Now fixed with a Makefile.

test/fixtures/AdapterTests.jl
FlorianFranzen: This will be addressed in a later update and is now tracked in #191.

test/fixtures/HostAPITests.jl
FlorianFranzen: This will be addressed in a later update and is now tracked in #191.

FlorianFranzen:
The remaining outstanding issues with the current testsuite are now tracked in #191. LGTM! :shipit:

test/adapters/substrate/adapter/wasm-adapter/Cargo.toml
drskalman: we don't test host api by calling them directly. we test them by running the wasm. So it is not an objection to the wasm-adapter name but the place of the adapter being in substrate subfolder. Could you add this also to #191 subtasks?

drskalman (approved):
I did some changes to the READMEs could you verify?

test/README.md
```diff
@@ -50,10 +52,6 @@ Needs recent version of Go.
 go build
 ```
 
-## On Debian-based systems
```
FlorianFranzen: This chapter is still Debian specific, so this header should stay here.

PR #189, opened 2020-05-06
Fix some inconsistencies with Host API spec

lamafab:
Some small adjustments of Host API function names. Can be easily reviewed with the Github diff tool.


lamafab:
@drskalman This takes like 30 seconds to review.

drskalman (approved):

Issue #195, opened 2020-05-19
Verify Grandpa Voting message definition and clarify Voter Set Id


drskalman:
The relevant questions from the implementor is ", it looks like some brackets are missing from the equation, could you clarify what is in the signature part of the message? why is stage included twice? also, could you clarify what V_id is: Vid: incremental change tracker counter"

drskalman:
- [ ] Add primary propose message to the grandpa spec
   - [ ] We always precommit to the ghost, if no ghost then to the base.
   - [ ] Replace the definition of prevoted block with prevote GHOST.
   - [ ] Fix Best-PreVote-Candidate.

drskalman:
- [X] Add primary propose message to the grandpa spec
   - [X] We always precommit to the ghost, if no ghost then to the base. SHouldn't ghost >= base be true.
   - [X] Replace the definition of prevoted block with prevote GHOST.
   - [X] Fix Best-PreVote-Candidate.

drskalman:
^ 7f46b4c - Best-Prevote-candidate is the GHOST - Example of non-finalized final candidate added.

drskalman:
Could @lamafab kindly takes care of Issue #130 now that he is at it?

lamafab:
Issues fixed in #202

drskalman:
f0a0b3fead92204fa830213e0b6c90d9e8188a4c :
 - Delay voting period in Grandpa statistically specified.
 - Honest scenario for best candidate not finalized added to grandpa example.

PR #202, opened 2020-05-20
Spec Authority set Id and Gossip Message

lamafab:
PR for issue #195. This also defines the "Gossip Message" type.

Also, please clarify "Finalizing Message" -> is this the `PrimaryPropose` in [this type](https://docs.rs/finality-grandpa/0.12.3/src/finality_grandpa/lib.rs.html#224-238)?

```rust
/// A protocol message or vote.
#[derive(Clone, PartialEq, Eq)]
#[cfg_attr(any(feature = "std", test), derive(Debug))]
#[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
pub enum Message<H, N> {
	/// A prevote message.
	#[cfg_attr(feature = "derive-codec", codec(index = "0"))]
	Prevote(Prevote<H, N>),
	/// A precommit message.
	#[cfg_attr(feature = "derive-codec", codec(index = "1"))]
	Precommit(Precommit<H, N>),
	/// A primary proposal message.
	#[cfg_attr(feature = "derive-codec", codec(index = "2"))]
	PrimaryPropose(PrimaryPropose<H, N>),
}
```

Printscreens:
image

image

image

### DIFF
--- a/polkadot-host-spec/c06-consensus.tm
+++ b/polkadot-host-spec/c06-consensus.tm
@@ -836,7 +836,8 @@
     amount of authority list (Definition <reference|defn-consensus-message-digest>)
     changes that occurred. Starting with the value of zero at genesis, the
     Polkadot Host increments this value by one every time a <strong|Scheduled
-    Change> or a <strong|Forced Change> occurs.
+    Change> or a <strong|Forced Change> occurs. The authority set Id is an
+    unsigned 64bit integer.
   </definition>
 
   <\definition>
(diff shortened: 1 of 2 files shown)


polkadot-host-spec/c06-consensus.tm
```diff
@@ -997,6 +1005,24 @@
 
   <subsection|GRANDPA Messages Specification>
 
+  <\definition>
+    <label|defn-gossip-message><strong|Gossip Message> is a variant as
```
drskalman: Is this Grandpa specific, i.e. we only have a Gossip Message for Granpa then better calling it GRANDPA gossip. If it is used somewhere else then better define it in the network section.

lamafab: Done!

PR #203, opened 2020-05-25
fix license headers

masterjedy:
Signed-off-by: Yuriy Zarudniy <zarudniy@soramitsu.co.jp>
Update kagome adapter license headers.


FlorianFranzen:
Thank you for taking care of this!

PR #206, opened 2020-05-28
Reorganize repo folder structure

FlorianFranzen:
As discussed on Riot I am proposing the following new folder structure. This also add makefiles for both spec and CI integration for the runtime spec.


lamafab (approved):

drskalman (changes requested):
All good except runtime folder is confusing as we also have runtime in tests. It could be chain-candidate-runtime or it can be chain-candidates/runtime in that case genesis should move there. Also genesis i think has spelling mistakes..

drskalman (approved):
:+1: LGTM

PR #207, opened 2020-06-01
Define and extend ShouldPropagate in algorithm 3.2

### LINKED ISSUE #180: Add broadcast transactions after validating them to the Host spec
We need to mandate in the Polkadot host spec that full node needs to broadcast all valid transactions:

the relevant part in substrate code:
 
"When we receive an extrinsic, that function is called: https://github.com/paritytech/substrate/blob/ee098e9e5c9ab813ff98a0fdb6dba25e1fad3ddf/client/network/src/protocol.rs#L1119

"We store, for the peer, that it knows the extrinsic and put it in the pool for validation
(known_extrinsics, for each peer, is an LRU cache)
   
Then later we gossip the validated ones: https://github.com/paritytech/substrate/blob/ee098e9e5c9ab813ff98a0fdb6dba25e1fad3ddf/client/network/src/protocol.rs#L1155
"

lamafab:
Fixes #180.

(ignore the branch name, it's only about propagation)

Printscreen:

image


polkadot-host-spec/c03-transition.tm
```diff
@@ -368,6 +375,12 @@
     <item><name|Maintain-Transaction-Pool> is described in Algorithm
     <reference|algo-maintain-transaction-pool>.
 
+    <item><name|ShouldPropagate> indictes whether the transaction should be
+    propagated based on the <verbatim|propagate> field in the
+    <verbatim|ValidTransaction> type as defined in Definition
+    <reference|defn-valid-transaction>, which is returned by
+    <verbatim|TaggedTransactionQueue_validate_transaction>.
+
```
drskalman, line 389: This does not make it clear what is the difference between "ShouldPropagate" and "Propagate" fields.

drskalman (changes requested):
- I would also add a general human friendly description at the beginning of the transaction section explaining the duty of the node in propagating transaction after validation refering this algorithm.
- I would also make it more clear which peer you should gossip the transaction at either here or in the transaction message. I assume we only propagate to people we are not sure they have seen the transaction.

lamafab:
@drskalman -> **Updated**

>     * I would also add a general human friendly description at the beginning of the transaction section explaining the duty of the node in propagating transaction after validation refering this algorithm.

image

>     * I would also make it more clear which peer you should gossip the transaction at either here or in the transaction message.

image

> I assume we only propagate to people we are not sure they have seen the transaction.

You're right:

```rust
	for (id, ref mut peer) in peers.iter_mut() {
		for (message_hash, topic, message) in messages.clone() {
			let intent = match intent {
				MessageIntent::Broadcast { .. } =>
					if peer.known_messages.contains(&message_hash) {
						continue;
					} else {
						MessageIntent::Broadcast
					},
				MessageIntent::PeriodicRebroadcast =>
					if peer.known_messages.contains(&message_hash) {
						MessageIntent::PeriodicRebroadcast
					} else {
						// peer doesn't know message, so the logic should treat it as an
						// initial broadcast.
						MessageIntent::Broadcast
					},
				other => other,
			};

			if !message_allowed(id, intent, &topic, &message) {
				continue;
			}

			peer.known_messages.insert(message_hash.clone());

			trace!(target: "gossip", "Propagating to {}: {:?}", id, message);
			network.write_notification(id.clone(), engine_id, message.clone());
		}
	}
```

polkadot-host-spec/c03-transition.tm
lamafab: The field `Propagate` (capitalized `propagate` -> I fixed it) refers to the data type returned by `TaggedTransactionQueue_validate_transaction`:

image

drskalman (approved):

PR #209, opened 2020-06-02
Spec inherents format and procedure for submitting 

### LINKED ISSUE #204: Specify the exact format that inhenets should be sent to BlockBuilder_inherent_extrinsics and the need to re-apply the inherents extrisics
points to deal with:

- [x] Input format for BlockBuilder_inherent_extrinsics
- [x] Which inherents should sent to BlockBuilder_inherent_extrinsics
- [x] Which inherents are expected to return.
- [x] the return inherents need to be re-applied again.

is it scaled sequence of Tuples or something else.

plus it seems that host need to call apply_extrinsics from the result of inherent_extrinsics but this is not seemed to be documented in the corresponding algorithm (in chapter 6) in the spec as well: (it assumes that runtime adds the inhenents when BlockBuilder_inherent_extrinsics which is not correct
   
related part of implementors conversation:

okay, it looked like it was storing something wrong. i did a fix and now i am seeing Timestamp must be updated once in the block, but I'm already calling inherent_extrinsics to set timstap0
Basti
No, that only returns the inherents as extrinsics that you need to apply with apply_extrinsic
lileth
oh i see, i'll try that, thanks!
lileth
Basti: inherent_extrinsics doesn't seem to be providing the right data, this is what i get back from it: [4 32 4 3 0 3 129 179 205 94] then decoded into an array of extrinsics becomes this: [[4 3 0 3 129 179 205 94]] when I try to call apply_extrinsic with that extrinsic it panics with Bad input data provided to apply_extrinsic. is there something I'm missing?
Heute
Basti
Your decoding looks wrong. A vector uses a compact u32 for storing the length and the value 1 should fit into one byte. But it looks like you skipped two bytes? 
lileth
i'm decoding into a slice of byte slices, so the 4 is the total length of the slice (1) and 32 is the length of the first byte slice (8)
unless inherent_extrinsics returns a different type? I thought it returned a vec of extrinsics
Basti
Ahh no that is right
I had forgotten that we have this length at the beginning of the extrinsic
But you prepend this 32 again when pushing it to the runtime?
lileth
ohh right that worked, thanks!
i also noticed, if i put timstap0 and babeslot into inherent_extrinsics, it only seems to return the timestamp to me, is that expected? i thought there were 4 inherents needed as per spec
or do i need to do all the inherents separately
Basti
Ignore
Yeah that is correct
babeslot will only be used on import when we call check_inherents
https://github.com/paritytech/substrate/blob/master/frame/babe/src/lib.rs#L647-L649
As you see here, we don't create an inherent

lamafab:
Fixes #204.

**Printscreens**
1.) Expand description of `BlockBuilder_inherent_extrinsics`
image

2.) Explain in more detail for "Inherents" actually are.
image

3.) Adjust algorithm 6.7 on how Inherents are generated and submitted.
image


lamafab:
**Updated**: correction of specifying the amount of key-value pairs for hashtables, rather than length.
image

polkadot-host-spec/ag-runtimeentries.tm
```diff
@@ -470,16 +470,22 @@
...
+  algorithm <reference|algo-build-block>.
 
   \;
 
   <strong|Arguments>:
 
   <\itemize>
     <item>A <name|Inherents-Data> structure as defined in
-    <reference|defn-inherent-data>.
+    <reference|defn-inherent-data>. This function does not generate an
```
drskalman: I'm not sure if it is a good place for this comment. Maybe at the return value because it is about what the function is generating? Related to that is that, should you or not submit the babeslot inherent to the function or not. If not, then how babeslot is being recorded in header then? In any case it is good to explain how babeslot is getting added to the block.

drskalman (changes requested):
Thanks a lot. Just one minor change and we should be able to merge.

drskalman (approved):
LGTM :+1:

PR #210, opened 2020-06-02
Revamp resulting type of `BlockBuilder_apply_extrinsic`

lamafab:
As discussed in the chat, this revamps the section of `BlockBuilder_apply_extrinsic`. Looks like there have been some adjustments (new error types) since the previous version was written a few months ago.

**Printscreen**
image
image

### DIFF
--- a/host-spec/ag-runtimeentries.tm
+++ b/host-spec/ag-runtimeentries.tm
@@ -420,7 +420,7 @@
 
   \;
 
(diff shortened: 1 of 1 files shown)


FlorianFranzen (commented):
The only thing I would change is the name of the "Appended" column. Maybe just rename it to "Type" or similar across all tables.

Otherwise a great job on getting that nested structure specified. This should be of great value to implementers.

PR #213, opened 2020-06-08
Import implementers guide

lamafab:
Imported the original [Implementer's Guide](https://github.com/paritytech/polkadot/blob/master/roadmap/implementors-guide/guide.md) into LaTex. Still needs some adjustments and polishing.


burdges:
It's under massive flux still, so maybe premature?

Issue #182, opened 2020-04-29
Adjustments based on child storage changes


lamafab:
Upstream changes of the child storage occurred. The following must be adjusted:

- [x] - Adjust spec (Host APIs)
- [x] - Adjust RPC PSP

**EDIT**
- [x] - Cleanup child storage section once #219 and #239 are merged.

FlorianFranzen:
After talking to @lamafab all of these were done already.

PR #219, opened 2020-06-09
Explain child storages

### LINKED ISSUE #212: The host spec should specify what child storage is
It also need to specifies its usage, example etc. 
Related conversation:
"
S: We have change trie section which discuss child storage: https://github.com/w3f/polkadot-spec/pull/142
It is also mentioned in appendix E I believe when we mention the related host api function I believe.
We have change trie section which discuss child storage: https://github.com/w3f/polkadot-spec/pull/142

P: Yeah but I have no idea what "the child storage" means"

lamafab:
**UPDATED**

As clarified in the Riot chat, Child Tries are now moved to Section 2.
image

FYI: This part is no longer relevant:
image

Most of this was removed/adjusted in #239. I'll do some additional cleanups once this PR and #239 are merged, tracked in -> #182.

**ORIGINAL**

Fixes #212.

Printscreen:

image


polkadot-host-spec/af-legacyhostapi.tm
```diff
@@ -20,6 +20,9 @@
 
   <section|Storage>
 
+  Interface for accessing the child storage from within the runtime. Child
+  Storages are defined in section <reference|sect-child-storages>.
+
```
drskalman: To me this implies that Storage interface is solely for accessing child storage.

polkadot-host-spec/ae-hostapi.tm
```diff
@@ -255,9 +255,21 @@
     containing the next key in lexicographic order.
   </itemize>
 
-  <section|Child Storage>
+  <section|Child Storage><label|sect-child-storages>
 
-  Interface for accessing the child storage from within the runtime.
+  Interface for accessing the child storage from within the runtime. Child
+  storages work the same way that regular storages do, but use two keys for
+  each value instead of just one. This allows for the functionality of having
```
drskalman: This is quite confusing. Perhapse 
1) make a difference in describing the child storage as a hashtable and the child storage trie. 
2) In describing the trie then one should clearly say how these two keys are going to make a trie. (As a new reader my first impression would be how a node can have two parents when there are two keys).
3) How the hash of the state trie is influenced by the child storage trie.

polkadot-host-spec/ae-hostapi.tm
```diff
@@ -255,9 +255,21 @@
...
+  identical keys with different values based on certain contextes, such as
+  separate storages for each Smart Contract and or Parachain, identified by
+  the preceding key. In case of Smart Contracts, the identifier could be an
+  address while for Parachain it could be the Parachain ID.
+
+  \;
+
+  Child storages are more performant than simply using prefixes, since the
```
drskalman: It s great that we have this comparison here but I think we can have it bit deeper. 
1) we can probably have example of a sub trie represented as prefix trie as well as a child trie and see how these two are diffenent.

polkadot-host-spec/ae-hostapi.tm
```diff
@@ -255,9 +255,21 @@
...
+  storage space can quickly be identified and does not require filtering the
+  entire database based on prefixes.
```
drskalman: One of the main part of requested clarification was to justify the need of child storage by some concrete example, could we have some here? (smar contract and how they are using child storage? is there any other example.)

drskalman (changes requested):

polkadot-host-spec/c02-state.tm
```diff
@@ -396,6 +396,42 @@
     </algorithmic>
   </algorithm>
 
+  <subsection|Child Tries><label|sect-child-trie-structure>
+
+  Child tries are trie structures as defined Section
```
drskalman: so you say
> [child tries] are saved as subcomponents within the Polkadot state trie.
1. what does subcomponent means?
then you say:
> Polkadot state trie only links to an individual child trie by its corresponding Merkle proof
2. What does links mean in the context of trie mathematically?
so because it is radix 16 trie you only need to tell me what is the (key,value) for the child trie then I can use radix 16 insertion to insert it in the trie
so my guess is that you mean value = hash of the child trie?
then what is the key?
3. as an implementor suppose 1. you embed the child trie into the state trie you are still compling with the spec because you are going to compute the correct merkle for the state trie, right?
so if 3 is correct, then why do we need to even mention the child trie, if it is just an optimization and doesn't change the final merkle hash value?

PR #232, opened 2020-06-26
Polkadot weight system review

drskalman:


lamafab:
@drskalman I don't think this way of writing/defining is necessary for the Polkadot Weight chapter:

image

The reader of "Polkadot Weights" does not need to implement certain component or mechanism in order to make their implementation compliant with the Polkadot network (like - for example - the Host spec or AnV). It's meant as a manual to understand the concept of Polkadot weights, why they're important, and how to assign those weights. Using definition and variables over-complicates this, in my opinion.

This paper is meant for **Runtime builders**, in the end they can use whatever weight system (and the corresponding weight values) that they want to.

lamafab:
See #255

Issue #208, opened 2020-06-02
Spec failure mode of ext_crypto_*_generate


FlorianFranzen:
Based on a question from Soramitsu: Specification says, that it should return simple i32 pointer to generated data, but what if it has to return error? e.g. key type id or the seed parameter can be invalid. What should it return in this case?

FlorianFranzen:
Substrate currently [panics on error](https://github.com/paritytech/substrate/blob/d65e6443c8281ff65fa5d39cbd454afa2624390e/primitives/io/src/lib.rs#L389).

lamafab:
Included in #239.

PR #239, opened 2020-07-09
Fix mistakes or oudated information in the spec

### LINKED ISSUE #217: Some mistakes or outdated information in the specs
Here is a list of mistakes or outdated information I've found in the specs:

- [x] Section G.2.2. mentions that `Core_execute_block` returns a boolean. It actually returns nothing. The runtime indicates a failure by trapping rather than returning `false`.

- [x] Section E.6.2 mentions that "The keys are little-endian, fixed-size integers". The keys are actually the SCALE-compact encoding of the index in the list. (https://github.com/paritytech/substrate/blob/b3e0cb68129eef29fd4521b23c39d1c8dd2e28cf/primitives/trie/src/lib.rs#L75) For example the key of the first element in the list is a SCALE-compact-encoded `0`.

- [x] The `ext_storage_append_version_1` function is missing from the specs. It expects as parameter a key and value (similar to `ext_storage_set`). It consists of three steps: load the storage item with the given key, assume that the existing storage item is a SCALE-encoded `Vec` and that the given value is a SCALE-encoded item of that vec, and add the given value to the end of that `Vec`. If the existing value is missing or is not actually an encoded `Vec`, store a one-length `Vec` with the given value.

- [x] This one might require messaging some people because I didn't manage to figure this out myself. The specs mentions that the `ext_storage_root_version_1` function returns the SCALE-encoded hash (so, one byte of length and 32 bytes with the hash itself). However, in practice, the runtime seems to expect the hash to be directly returned (32 bytes, without any length prefix). The weird thing is that it seems that Substrate's source code agrees with the hash being SCALE-encoded. I haven't tried the `ext_storage_changes_root_version_1` function, but it might be the same.

- [x] The `ext_crypto_start_batch_verify_version_1`, `ext_crypto_sr25519_verify_version_2` and `ext_crypto_start_batch_verify_version_1` functions are missing from the specs. Many other functions are missing, but these three are used basically all the time by the runtime.

- [x] Definition 2.11 mentions `scale_encode(storedvalue(kn))`. Definition 2.1 mentions that `storedvalue` is empty if the node doesn't have any storage value. This has lead me to believe that `scale_encode(storedvalue(kn))`, when `kn` has no storage value, should be one byte of value `0`. In reality it should be nothing.

- [x] Definition 2.11 defines `svn` as bytes. In my opinion it should clearly mention that `childrenbitmap` should be the little-endian encoding of the value defined in definition 2.10. In other words, the first byte of `svn` should be `b7...b0` and the second byte of `svn` should be `b15...b8`.

- [x] Section G.2.1. is missing the `transaction_version: u32` field that was recently added: https://github.com/paritytech/substrate/blob/dbf2163250833e6ac898e7f6c3c8f89f08a7c19d/primitives/version/src/lib.rs#L107

- [x] Section G.2.1. mentions `ApisVec`, but this isn't defined anywhere (it's a `Vec<([u8; 8], u32)>`).

- [x] Section E.1.3. says that `offset` is a `i32`. It is in fact a `u32`. Negative offsets don't really make sense anyway.

lamafab:
Fixes #217.

**UPDATED**: The following, missing APIs are now included:

- ext_storage_append
- ext_storage_start_transaction
- ext_storage_rollback_transaction
- ext_storage_commit_transaction
- ext_crypto_ecdsa_public_keys
- ext_crypto_ecdsa_generate
- ext_crypto_ecdsa_sign
- ext_crypto_ecdsa_verify
- ext_crypto_start_batch_verify
- ext_crypto_finish_batch_verify
- ext_trie_keccak_256_root
- ext_trie_keccak_256_ordered_root

The Child Storage APIs are adjusted (and renamed) to upstream changes:

- ext_default_child_storage_set
- ext_default_child_storage_get
- ext_default_child_storage_read
- ext_default_child_storage_clear
- ext_default_child_storage_storage_kill
- ext_default_child_storage_exists
- ext_default_child_storage_clear_prefix
- ext_default_child_storage_root
- ext_default_child_storage_next_key

Offchain API are still open and will be tracked in another ticket (#241).

**Printscreens**

>  Section G.2.2. mentions that Core_execute_block returns a boolean. It actually returns nothing. The runtime indicates a failure by trapping rather than returning false.

image
image

>  Section E.6.2 mentions that "The keys are little-endian, fixed-size integers". The keys are actually the SCALE-compact encoding of the index in the list. (https://github.com/paritytech/substrate/blob/b3e0cb68129eef29fd4521b23c39d1c8dd2e28cf/primitives/trie/src/lib.rs#L75) For example the key of the first element in the list is a SCALE-compact-encoded 0.

image

>  This one might require messaging some people because I didn't manage to figure this out myself. The specs mentions that the `ext_storage_root_version_1` function returns the SCALE-encoded hash (so, one byte of length and 32 bytes with the hash itself). However, in practice, the runtime seems to expect the hash to be directly returned (32 bytes, without any length prefix). The weird thing is that it seems that Substrate's source code agrees with the hash being SCALE-encoded. I haven't tried the `ext_storage_changes_root_version_1` function, but it might be the same.

image
image

> The `ext_storage_append_version_1` function is missing from the specs. It expects as parameter a key and value (similar to `ext_storage_set`). It consists of three steps: load the storage item with the given key, assume that the existing storage item is a SCALE-encoded Vec and that the given value is a SCALE-encoded item of that vec, and add the given value to the end of that Vec. If the existing value is missing or is not actually an encoded Vec, store a one-length Vec with the given value.

image

>  The `ext_crypto_start_batch_verify_version_1`, `ext_crypto_sr25519_verify_version_2` and `ext_crypto_start_batch_verify_version_1` functions are missing from the specs. Many other functions are missing, but these three are used basically all the time by the runtime.

(`ext_crypto_sr25519_verify_version_2` was already specified. Maybe Pierre missed it.)

image
image

image
image

I also specced the new ECDSA functions.

image
image

>  Definition 2.11 mentions `scale_encode(storedvalue(kn))`. Definition 2.1 mentions that `storedvalue` is empty if the node doesn't have any storage value. This has lead me to believe that `scale_encode(storedvalue(kn))`, when `kn` has no storage value, should be one byte of value 0. In reality it should be nothing.

>  Definition 2.11 defines `svn` as bytes. In my opinion it should clearly mention that `childrenbitmap` should be the little-endian encoding of the value defined in definition 2.10. In other words, the first byte of `svn` should be `b7...b0` and the second byte of svn should be `b15...b8`.

image

>  Section G.2.1. is missing the `transaction_version: u32` field that was recently added: https://github.com/paritytech/substrate/blob/dbf2163250833e6ac898e7f6c3c8f89f08a7c19d/primitives/version/src/lib.rs#L107
>  Section G.2.1. mentions ApisVec, but this isn't defined anywhere (it's a `Vec<([u8; 8], u32)>`).

image

>  Section E.1.3. says that `offset` is a i32. It is in fact a u32. Negative offsets don't really make sense anyway.

image
image
image


FlorianFranzen:
>This one might require messaging some people because I didn't manage to figure this out myself. The specs mentions that the ext_storage_root_version_1 function returns the SCALE-encoded hash (so, one byte of length and 32 bytes with the hash itself). However, in practice, the runtime seems to expect the hash to be directly returned (32 bytes, without any length prefix). The weird thing is that it seems that Substrate's source code agrees with the hash being SCALE-encoded. I haven't tried the ext_storage_changes_root_version_1 function, but it might be the same.

As described in definition E.2 in the spec, fixed length data does not include any length information. The length of the hash is always known. Gossamer for example just encodes [hashes and fixed byte arrays without prefix](https://github.com/ChainSafe/gossamer/blob/development/lib/scale/encode.go#L75-L78). 

In Substrate it is returned as `H256` which is passed as `[u8, 32]` as you can see [here](https://github.com/paritytech/substrate/blob/master/primitives/runtime-interface/src/impls.rs#L372-L403). And according to the [translation table](https://github.com/paritytech/substrate/blob/master/primitives/runtime-interface/src/lib.rs#L79-L103)  `[u8, 32]` is encoded as 32bit pointer.

FlorianFranzen:
> Section G.2.2. mentions that Core_execute_block returns a boolean. It actually returns nothing. The runtime indicates a failure by trapping rather than returning false.

I would still document the fact that it panics on failure. You could even copy that sentence a few times and fix #208 as well.

FlorianFranzen (approved):
Looks great. Thank you for the thorough update.

PR #242, opened 2020-07-14
Polkadot weight system rational and computation

drskalman:


drskalman:
There was another pull request already.

PR #250, opened 2020-07-23
Spec transaction format

lamafab:
Hi

As discussed and partially already reviewed by @FlorianFranzen, here are the remaining changes. I still do not fully understand how the mortality of a transaction is evaluated on the side of the Runtime. But in this context (as requested by other teams) it's mostly important how to **create** a transactions, not necessarily validate it. But we should follow up on this and not merge before we know all the details.

@FlorianFranzen maybe you can figure out what is going on here?

`CheckMortality` makes it into the transaction by encoding `Era` (`PhantomData` is skipped). `Era` is essentially `period` + `phase`. This part is specced.

```rust
// in `substrate/frame/system/src/extensions/check_mortality.rs`

/// Check for transaction mortality.
#[derive(Encode, Decode, Clone, Eq, PartialEq)]
pub struct CheckMortality<T: Trait + Send + Sync>(Era, sp_std::marker::PhantomData<T>);
```

```rust
// in `substrate/primitives/runtime/src/generic/era.rs`

impl Era {
	pub fn mortal(period: u64, current: u64) -> Self {
		let period = period.checked_next_power_of_two()
			.unwrap_or(1 << 16)
			.max(4)
			.min(1 << 16);
		let phase = current % period;
		let quantize_factor = (period >> 12).max(1);
		let quantized_phase = phase / quantize_factor * quantize_factor;

		Era::Mortal(period, quantized_phase)
	}

         // ...
}

impl Encode for Era {
	fn encode_to<T: Output>(&self, output: &mut T) {
		match self {
			Era::Immortal => output.push_byte(0),
			Era::Mortal(period, phase) => {
				let quantize_factor = (*period as u64 >> 12).max(1);
				let encoded = (period.trailing_zeros() - 1).max(1).min(15) as u16 | ((phase / quantize_factor) << 4) as u16;
				output.push(&encoded);
			}
		}
	}
}

impl codec::EncodeLike for Era {}

impl Decode for Era {
	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
		let first = input.read_byte()?;
		if first == 0 {
			Ok(Era::Immortal)
		} else {
			let encoded = first as u64 + ((input.read_byte()? as u64) << 8);
			let period = 2 << (encoded % (1 << 4));
			let quantize_factor = (period >> 12).max(1);
			let phase = (encoded >> 4) * quantize_factor;
			if period >= 4 && phase < period {
				Ok(Era::Mortal(period, phase))
			} else {
				Err("Invalid period and phase".into())
			}
		}
	}
}
```

The transaction author includes the block hash which starts the mortality (`current` parameter in `Era::mortal`) in the payload, which will be signed.

```rust
        // in `substrate/frame/system/src/extensions/check_mortality.rs`
	fn additional_signed(&self) -> Result<Self::AdditionalSigned, TransactionValidityError> {
		let current_u64 = <Module<T>>::block_number().saturated_into::<u64>();
		let n = self.0.birth(current_u64).saturated_into::<T::BlockNumber>();
		if !<BlockHash<T>>::contains_key(n) {
			Err(InvalidTransaction::AncientBirthBlock.into())
		} else {
			Ok(<Module<T>>::block_hash(n))
		}
	}
```

After the transaction is sent, the validators can *somehow* figure out which block hash is included in the payload just with the values `period` and `phase`. The validators has to be able to reproduce the payload in order to verify the signature.

Obviously, this figuring-out part is done with `CheckMortality::validate()`, but it's not clear how this will select the correct block. This looks like trial and error to me..??


runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+
+    \[
+        tx := (T_v, T_b)
+    \]
+
+    The value of $T_b$ varies for each version. The current version 4 is
+    described in section \ref{sect-version-four}.
+\end{definition}
+
+\begin{definition}
```
FlorianFranzen, line 25: IMHO I would not mention the binary operations at all but just say that `T_v` is a bit field, where the first bit signals if the transaction is signed and the remaining 7 bit represent the version number from 0 to 127.

runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+        0, & \text{Ed25519, followed by: } (b_0, ...,b_{63}) \\
+        1, & \text{Sr25519, followed by: } (b_0, ...,b_{63}) \\
+        2, & \text{Ecdsa, followed by: } (b_0, ...,b_{64})
+        \end{cases}
+    \]
+
+    Signature types vary in sizes, but each individual type is always fixed-size
+    and therefore does not contain a length prefix. \verb|Ed25519| and
+    \verb|Sr25519| signatures are 512-bit while \verb|Ecdsa| is 560-bit, where
+    the last 8 bits are the recovery ID.
```
FlorianFranzen, line 85: This is confusing to me: The Ecdsa signature is 560 bit = 70 byte, but the definition only accounts for 65 bytes = 520 bit.

runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+                \end{flalign*}
+            }$
+        \end{algorithmic}
+    \end{algorithm}
+
+    \begin{itemize}
+        \item $T^{b0}_{mort}$: the first byte of $T_m$.
+        \item $T^{b1}_{mort}$: the second byte of $T_m$.
+        \item {\sc Min(..)}: returns the minimum of two values.
+        \item {\sc Max(..)}: returns the maximum of two values.
```
FlorianFranzen: It could potentially improve readability if you replace the `min` and `max` applications by something like `limit` or similar, because in the end this is why they are used here.

E.g. if you want to make sure that `1 <= x <= 10`, you can use `min(max(x, 1), 10)`, but it becomes much more readable if you use `limit(x, 1, 10)` or instead.

runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+            \Statex 
+            \State \textbf{Init} $enc = T^{b0}_{mort} + (T^{b1}_{mort} << 8)$
+            \State \textbf{Init} $M_{per} = 2 << (enc\ mod\ (1 << 4))$
+            \State \textbf{Init} $factor = \textsc{Max}(M_{per} >> 12, 1)$
+            \State \textbf{Init} $M_{pha} = (enc >> 4) * factor$
+            \State $\vcenter{
+                \begin{flalign*}
+                    \Return & 
+                    \begin{cases}
+                    (M_{per}, M_{pha}) & if\ M_{per} \geq 4\ \textbf{and}\ M_{pha} < M_{per} \\
```
FlorianFranzen: It is probably in the code like this, but I would probably move these requirements up into the definition. That the phase is smaller then the period should is a hard requirement. And the encoding is not able to encode periods < 4 either.

runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+        such as \verb|32|, \verb|64|, \verb|128|, etc.
+        \item $M_{pha}$: the phase in the period that this extrinsic's lifetime
+        begins. This value is calculated with a formula and validators can use
+        this value in order to determine which block hash is included in the
+        payload.
+    \end{itemize}
+
+    The extrinsic author uses the number of the block which starts the mortality
+    period, $H_i(B)$, and specifies the desired period in order the calculate
+    the phase:
```
FlorianFranzen: You could also say, that in order to tie a transaction's lifetime to a certain block after they were issued, without wasting precious space for block hashes, block numbers are divided into regular periods and the lifetime is instead expressed as a "phase" from these regular boundaries.

runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+                \begin{flalign*}
+                    \Return & 
+                    \begin{cases}
+                    \textit{Immortal} & if\ T^{b0}_{mort} = 0
+                    \end{cases}&
+                \end{flalign*}
+            }$
+            \Statex 
+            \State \textbf{Init} $enc = T^{b0}_{mort} + (T^{b1}_{mort} << 8)$
+            \State \textbf{Init} $M_{per} = 2 << (enc\ mod\ (1 << 4))$
```
FlorianFranzen, line 235: You could also just say that the last 4 bit encode the period: If the last 4 bits are x \elem [1, 15], then the period = 2^{x + 1}. As a result the period \elem [2^2, 2^16].

runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+            \State \textbf{Init} $factor = \textsc{Max}(M_{per} >> 12, 1)$
```
FlorianFranzen: The period is encoded as a power, but the phase is not. So whenever the period > 2^12, as the phase is limited to 12 bit precision, the phase is corrected by a factor:

 - phase <= 2^12 => factor = 1
 - phase  = 2^13 => factor = 2^13 / 2^12 = 2^1 = 2
 - phase = 2^14 => factor = 2^14 / 2^12 = 2^2 = 4
 - phase = 2^15 => factor = ... = 8
 - phase = 2^16 => factor = ... = 16

In the end that just means, the phase encoding loses precision, but is at least able to express approximate values across the full period.

runtime-spec/runtime-spec-transaction.tex
```diff
@@ -0,0 +1,333 @@
...
+            \State \textbf{Init} $M_{pha} = (enc >> 4) * factor$
```
FlorianFranzen, line 237: The phase is encoded in the top 12 bit and correct by a factor if the phase is larger then the periods 12-bit precision (see comment above).

runtime-spec/runtime-spec-transaction.tex
FlorianFranzen: It is probably in the code like this, but I would move these requirements up into the definition. That the phase is smaller then the period should is a hard requirement. And the encoding is not able to encode periods < 4 either.

FlorianFranzen (changes requested):
Very nice write-up, so minor remarks, mostly about my love for bit-fields. Also added a bit of info how I think mortality is encoded.

runtime-spec/runtime-spec-transaction.tex
lamafab: @FlorianFranzen I adjusted the rest, but left this (and https://github.com/w3f/polkadot-spec/pull/250#discussion_r464954623) open. Do do you think this explanation here should be put inside the document?

(thanks for this, btw)

PR #253, opened 2020-07-28
Fix and improve kagome buil

xDimon:
Fix of build kagome over hunter package manager.
Activate hunters binary cache for accelerate building.

Signed-off-by: Dmitriy Khaustov aka xDimon <khaustov.dm@gmail.com>

### DIFF
--- a/test/adapters/kagome/cmake/HunterConfig.cmake
+++ b/test/adapters/kagome/cmake/HunterConfig.cmake
@@ -17,6 +17,11 @@
 
 # TODO: Move away from custom kagome branch for testing
 hunter_config(kagome
-    URL https://github.com/soramitsu/kagome/archive/af5c718a05a3b734291dd2c4f574c8dffb76ad46.tar.gz
-    SHA1 81f456bb92d06e7b5861dacb66d4ff77515dadf6
+    URL https://github.com/soramitsu/kagome/archive/219d46ff77ae8b77974373653ce25e720ef0645f.tar.gz
+    SHA1 1a0827b729b4dc1e40e37887df53f05bbcac3607
 )
+hunter_config(libsecp256k1
+    URL https://github.com/soramitsu/soramitsu-libsecp256k1/archive/c7630e1bac638c0f16ee66d4dce7b5c49eecbaa5.zip
+    SHA1 179e316b0fe5150f1b05ca70ec2ac1490fe2cb3b
+    CMAKE_ARGS SECP256K1_BUILD_TEST=OFF
+    )
--- a/test/adapters/kagome/cmake/HunterInit.cmake
+++ b/test/adapters/kagome/cmake/HunterInit.cmake
@@ -1,0 +1,31 @@
+# specify GITHUB_HUNTER_TOKEN and GITHUB_HUNTER_USERNAME to automatically upload binary cache to github.com/soramitsu/hunter-binary-cache
+# https://docs.hunter.sh/en/latest/user-guides/hunter-user/github-cache-server.html
+string(COMPARE EQUAL "$ENV{GITHUB_HUNTER_TOKEN}" "" password_is_empty)
+string(COMPARE EQUAL "$ENV{GITHUB_HUNTER_USERNAME}" "" username_is_empty)
+
+# binary cache can be uploaded to soramitsu/hunter-binary-cache so others will not build same dependencies twice
+if (NOT password_is_empty AND NOT username_is_empty)
+    option(HUNTER_RUN_UPLOAD "Upload cache binaries" YES)
+    message("Binary cache uploading is ENABLED.")
+else ()
+    option(HUNTER_RUN_UPLOAD "Upload cache binaries" NO)
+    message("Binary cache uploading is DISABLED.")
+endif ()
+
+set(
+    HUNTER_PASSWORDS_PATH "${CMAKE_CURRENT_LIST_DIR}/HunterPasswords.cmake"
+    CACHE FILEPATH "Hunter passwords"
+)
+
+set(
+    HUNTER_CACHE_SERVERS "https://github.com/soramitsu/hunter-binary-cache;"
+    CACHE STRING "Binary cache server"
+)
+
+include(${CMAKE_CURRENT_LIST_DIR}/HunterGate.cmake)
+
+HunterGate(
+    URL https://github.com/soramitsu/soramitsu-hunter/archive/v0.23.254-soramitsu1.tar.gz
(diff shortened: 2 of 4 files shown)


test/adapters/kagome/cmake/HunterConfig.cmake
```diff
@@ -17,6 +17,11 @@
 
 # TODO: Move away from custom kagome branch for testing
 hunter_config(kagome
-    URL https://github.com/soramitsu/kagome/archive/af5c718a05a3b734291dd2c4f574c8dffb76ad46.tar.gz
-    SHA1 81f456bb92d06e7b5861dacb66d4ff77515dadf6
+    URL https://github.com/soramitsu/kagome/archive/219d46ff77ae8b77974373653ce25e720ef0645f.tar.gz
+    SHA1 1a0827b729b4dc1e40e37887df53f05bbcac3607
```
FlorianFranzen, line 21: Do we still rely on a custom branch or is the version in `soramitsu-hunter` just to outdated?

test/adapters/kagome/cmake/HunterInit.cmake
```diff
@@ -0,0 +1,31 @@
...
+else ()
+    option(HUNTER_RUN_UPLOAD "Upload cache binaries" NO)
+    message("Binary cache uploading is DISABLED.")
+endif ()
+
+set(
+    HUNTER_PASSWORDS_PATH "${CMAKE_CURRENT_LIST_DIR}/HunterPasswords.cmake"
+    CACHE FILEPATH "Hunter passwords"
+)
+
```
FlorianFranzen, line 19: Anything above those lines is not really relevant for use and based on confusing variable names only used on your CI. This does not belong upstream and can all be easily set over some environmental variables.

test/adapters/kagome/cmake/HunterPasswords.cmake
```diff
@@ -0,0 +1,11 @@
...
+  # REPO_OWNER + REPO = https://github.com/forexample/hunter-cache
+  REPO_OWNER "soramitsu"
+  REPO "hunter-binary-cache"
+
+  # USERNAME = warchant
+  USERNAME "$ENV{GITHUB_HUNTER_USERNAME}"
+
+  # PASSWORD = GitHub token saved as a secure environment variable
+  PASSWORD "$ENV{GITHUB_HUNTER_TOKEN}"
+)
```
FlorianFranzen, line 11: Same here. And we have no control over this repo. I am sure hunter can be configured in a similar fashion just [with  environmental variables](https://hunter.readthedocs.io/en/latest/reference/user-variables.html) to set this directly with a few exports and no upstream changes.

test/adapters/kagome/cmake/HunterConfig.cmake
```diff
@@ -17,6 +17,11 @@
...
 )
+hunter_config(libsecp256k1
+    URL https://github.com/soramitsu/soramitsu-libsecp256k1/archive/c7630e1bac638c0f16ee66d4dce7b5c49eecbaa5.zip
+    SHA1 179e316b0fe5150f1b05ca70ec2ac1490fe2cb3b
+    CMAKE_ARGS SECP256K1_BUILD_TEST=OFF
+    )
```
FlorianFranzen, line 27: This should not be required as it is not a direct dependency.

FlorianFranzen (changes requested):
I am more then willing to update hunter and kagome, as well as adding the link to the binary cache. However I would prefer not to include any of the upload related code, as it is highly specific to how you use hunter and could be just as well handled through setting a few environmental variables.

test/adapters/kagome/cmake/HunterConfig.cmake
FlorianFranzen: Ok, just took a better look. It seems to be a recent branch of master with single commit that adds a version to the CMake file. Will that be merged into master?

FlorianFranzen:
### Current CI failures
```
CMake Error at CMakeLists.txt:56 (add_executable):
  Target "kagome-adapter" links to target "kagome::polkadot_trie_batch" but
  the target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?

CMake Error at CMakeLists.txt:56 (add_executable):
  Target "kagome-adapter" links to target "kagome::trie_db_backend" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
```
Yes, this is a pet peeves of mine. The definition of `kagome_LIBRARIES` you export in your package config is hard coded and poorly updated (e.g. on nix I use [this patch](https://gitlab.w3f.tech/florian/w3fpkgs/-/blob/master/pkgs/kagome/kagome_LIBRARIES.patch).). This should be easy to fix with an automatically generate variable. 

```
CMake Error at CMakeLists.txt:56 (add_executable):
  Target "kagome-adapter" links to target "secp256k1::secp256k1" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?

CMake Error at CMakeLists.txt:56 (add_executable):
  Target "kagome-adapter" links to target "Boost::filesystem" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```
Another bug in the package config, which fails to check for its propagated dependencies with `add_dependencies` (as recommended in the [official CMake docs](https://cmake.org/cmake/help/v3.17/manual/cmake-packages.7.html#creating-a-package-configuration-file) for details). We currently do all of that for kagome [in our CMakeLists.txt](https://github.com/w3f/polkadot-spec/blob/master/test/adapters/kagome/CMakeLists.txt#L49-L58), but it breaks every time you add a dependency (or even just a new boost module).

FlorianFranzen:
Even if I fix the build failures mentioned above, the newer version of kagome breaks compilation further, including but not limited to:
  - `kagome::runtime::SizeType` is now `WasmSize`
  - `kagome::storage::trie` was refactored.
  - `kagome::extensions` API was extended.

Did you test this before submission?

FlorianFranzen:
This will be replaced by #256. I was able to fix all aforementioned issues and included three of your requested changes in 0df699b, 8e3d81f and 889772e. 

You can now run `make enable-bincache` inside the adapter directory to have hunter use the suggested binary cache with these changes in the future.

Thank you for your contribution.

test/adapters/kagome/cmake/HunterConfig.cmake
FlorianFranzen: Nevermind, my bad. I had to find out the hard way that _libsecp256k1_ otherwise misses a `VERSION` field.

PR #262, opened 2020-08-13
wasm-adapter: Fix child storage root shim

FlorianFranzen:
This fixes a typo in the child storage root test function in the wasm-adapter runtime. However it seems our testsuite was testing for this, so it will need to be updated as well.


lamafab (approved):
LGTM

PR #264, opened 2020-08-18
Update Gossamer's Host API testsuite integration 

FlorianFranzen:
This adds legacy Host API test support to the gossamer-adapter. It required the removal of some unused Host API functions from the wasm adapter as well the extension of the testsuite to provide libwasmer.so to the gossamer adapter, as it currently can not be linked statically.


FlorianFranzen:
The two remaining falling test were moved to #295.

lamafab (approved):

PR #269, opened 2020-08-30
kagome-adapter: implement host-api storage next-key test 

FlorianFranzen:
This implements the test of `ext_storage_next_key_version_1` host-api for the `kagome-adapter`. It currently fails, as Kagome assumes the passed key has to be a valid, which causes it to panic. In reality it should just return the next lexicographic key in storage or an empty result if there are not further keys after the passed value in storage. 

Before this can be merged the following things have to happen:
 - [x] The spec has to be updated to reflect that the key does not have exists (#272)
 - [x] Fix double encoding in wasm adapter for next_key and get (1d94faf)
 - [x] Update kagome- and substrate-adapter to fixed encoding  (f0341d3, a1d7515, 5af6362) 
 - [x] Depends on Kagome fixing `storage_next_key`, see soramitsu/kagome#510 (not in master, using branch under review)
 - [x] Depends on Kagome merging our fix for `storage_get`, see soramitsu/kagome#518

PR #272, opened 2020-08-31
Define lexicographic ordering

lamafab:
Fixes checklist item 1.) of #269.

Printscreens:

image
image
image


FlorianFranzen (approved):
Looks good. You might want to include the updated PDF though.

Issue #261, opened 2020-08-13
Prepare a HackMD document for libp2p interoperability problem


drskalman:
The document should describe the problem, demonstrate discrepancy between implementation, steps we took to involve the upstream, and our plan of action due to lack of interest from them.

drskalman:
@Noc2 could you simply start the hackmd, leave the link here and then re-assign the issue to @FlorianFranzen to start the text.

drskalman:
I forgot I promised to make the hackmd myslef :-P

drskalman:
Here is the hackmd.  https://hackmd.io/L0TfbtPoQ-yrc0xSRHNazA?both

lamafab:
@FlorianFranzen will talk with Soramitsu, @infinity0 and ChainSafe to figure out the interoperability problems and write the required spec in order to fix those.

lamafab:
@lamafab @FlorianFranzen  organize a meeting with @infinity0 and discuss this.

PR #276, opened 2020-09-02
Networking spec update

lamafab:
@FlorianFranzen 

Incomplete version of the networking spec, currently blocked. See https://github.com/w3f/polkadot-spec/issues/261.

Networking PDF: https://github.com/w3f/polkadot-spec/blob/spec-networking/runtime-spec/networking-spec.pdf


FlorianFranzen (changes requested):
Please move everythinh into the host-spec folder, then we can merge it for now. I will do the CI in a separate PR.

FlorianFranzen (approved):

PR #278, opened 2020-09-09
kagome-adapter: Use correct return type for ext_storage_get

FlorianFranzen:
This now uses the correct return type for `ext_storage_get_version_1`. 

We should be merged this as soon as upstream Kagome fixes the underlying issue of encoding the return value correctly as tracked in soramitsu/kagome#518.

Based on top of #269, will need to be merged first.


FlorianFranzen:
Merged into #269.

FlorianFranzen:
Current failure described [here](https://github.com/soramitsu/kagome/pull/510#issuecomment-689024278).

PR #295, opened 2020-10-07
gossamer-adapter: Add missing Host API hashing functions

FlorianFranzen:
Adds the last two missing hashing function to the `gossamer-adapter`:

 - Fix `ext_keccak_256` test
 - Fix `ext_blake2_128` test
 - Support `wasmer` and `wasmtime` runtime backends


FlorianFranzen:
Even after switching the runtime environment from wasmer to wasmtime, the hashes are still incorrect. The switch also broke `test_twox_64` again.

Now build against Gossamer from ChainSafe/gossamer#1127.

FlorianFranzen:
Fixed underlying broken hashing implementation in ChainSafe/gossamer#1131. 

Added a flag to the adapter to switch the wasm environment to wasmtime (`--wasmtime`) instead of making it the default.

Now passes all test with `wasmer` and fix for `wasmtime` available in ChainSafe/gossamer#1131.

lamafab (approved):

PR #296, opened 2020-10-09
gossamer-adapter: Add set_get_storage test

FlorianFranzen:
This is how Gossamer's Host API currently returns stored values:

| key | value | result |
| ---- | -------- | ------- |
| static | Inverse | statice |
| function | Horizontal |  functionl |
| Integrated | portal |  portal |
| non-based | Monitored | non-based |
| productivity | secondary | productiv |
| Exclusive | next generation | Exclusiveration |
| disintermediate | Grass-roots | disintermed |
| contingency | value-added | contingency |
| human-resource | Reactive | Reactive |
| Optional | secondary | Option |

Result is the same on wasmer and new wasmtime backend (use `--wasmtime` flag to test).

It looks like there is clearly some kind of order to it, so expect there to be a bug in the storage itself at this point.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -170,6 +170,31 @@ jobs:
         name: gossamer-adapter
         path: test/lib/libwasmer.so
 
+  build-adapter-gossamer-legacy:
+    name: "[build] gossamer-adapter-legacy"
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Cache go modules
+      uses: actions/cache@v1
+      with:
+        path: ~/go/pkg/mod
+        key: go-mod-adapter-gossamer-legacy-${{ hashFiles('test/adapters/gossamer-legacy/go.sum') }}
+        restore-keys: go-mod-adapter-gossamer-legacy-
+    - name: Build gossamer legacy adapter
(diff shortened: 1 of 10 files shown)


syrusakbary:
Sad to see that Wasmer appeared broken for you. Our 0.X versions were not as polished as they should have been.

We have been working in the last 8 months on a complete refactor that fixes most of the issues with a much better architecture: Wasmer 1.0 (we are in [alpha 4 as of now](https://github.com/wasmerio/wasmer)!).

We are working on upgrading our [Wasmer-Go integration at the moment](https://github.com/wasmerio/wasmer-go/compare/master...Hywan:reboot?expand=1) (moving from 0.17 to 1.0.0-alpha)... would love to hear if you are having any issues with it!

FlorianFranzen:
@syrusakbary Thank you for your update. We are excited to see progress on the wasmer refactor and its go bindings. Furthermore, for this particular issue at least, the failure seems to be unrelated to wasmer itself as I can reproduce it with Gossamer's wasmtime backend as well.

drskalman:
the last changes that @FlorianFranzen has made to gossamer code makes the test pass but still the bug need to be investigated on the chainsafe side. So we wait for @noot to fix the bug on gosammer side before merging it now.

FlorianFranzen:
The fix was merged upstream and ChainSafe regard the issue as solved. We will leave it at that for now. Will merge this as soon as the CI passes.

PR #297, opened 2020-10-10
update gossamer adapter to use wasmtime

noot:
This updates the gossamer adapter to work with the updated gossamer runtime package API. It also updates the hashing tests to use the wasmtime adapter. However this can easily be switched out for wasmer by using the `"github.com/ChainSafe/gossamer/lib/runtime/wasmer"` package instead. Unfortunately this does not fix the issue with `blake2_128` and `keccak_256`, will continue to investigate it.


FlorianFranzen:
Thank you for update. Very much appreciated!

PR #298, opened 2020-10-10
gossamer-adapter: update to use wasmtime

FlorianFranzen:
This switches the runtime environment for the Host API tests from wasmer to wasmtime.

Currently break `test_twox_64` which now only returns all-zero hashes.

Build against Gossamer from ChainSafe/gossamer#1127.


FlorianFranzen:
As the switch did not fix the storage (or hashing) issues, there is no need to merge this separately. I will just include in #295 as initially planned.

PR #317, opened 2020-11-05
hostapi-runtime: use imported instead of exported memory

FlorianFranzen:
This should bring the hostapi runtime closer to what main net runtimes are doing. Will need some additional work on each adapter and potentially upstream as well.


FlorianFranzen:
This is now part of #351.

PR #320, opened 2020-11-11
Fix wrong usage of extrinsics

### LINKED ISSUE #27: Correct the problem with calling mistakenly transaction, extrinsic in the extrinsics section
When the Extrinsics section is written with the confusion that the inhernts can be gossiped on the 
network this is not true however, So in many places "Extrinsic" can be replaced by "Transaction".

drskalman:
closes #27 
@lamafab as you have written the transaction spec, I'd appreciate to check the accuracy of the section.
@FlorianFranzen I changed the paper type from papyrus to paper but the margines are still messed up due to std-latex package. Could you take a look.


FlorianFranzen (approved):
It seems you already remove `std-latex`, so looks good to me.

lamafab (approved):

FlorianFranzen:
Rebased commits on lastest master. Ready to merge.

Issue #22, opened 2019-08-29
Update the Trie Spec to branch value should be before the child value (after bitmap)


drskalman:
https://github.com/paritytech/substrate/blob/6d949bca3c9ba74ce4b64928e965930758044b53/core/trie/src/node_codec.rs#L160 would be for reference code, 
simply kept the order of the former substrate encoding.
we need to move the value 
after the bitmap and before the successive children
(bitmap bytes are reserved and written after)

drskalman:
Assigned to @lamafab for review.

PR #326, opened 2020-12-02
close issue #22 by movin the branch node value before the its children…

drskalman:
… hashes, I checked the rust code and it seems to me that it agree with definition 2.11 but if @lamafab can double check that before merging would be awesome. The rust code is pointed at in issue #22


lamafab:
@lamafab will review this.

lamafab (approved):

PR #338, opened 2021-01-07
testsuite: use submodules and update dependencies

FlorianFranzen:
To simplify the update of the implementations the testsuite is run against, this PR now builds the kagome and gossamer adapter against the host submodule (again).

This however is not easily archived for substrate, as the polkadot host and the underlying substrate code are in separate repository, requiring manual synchronization. The fact that polkadot is build against substrate master also does not make this easier.

As a result the substrate-adapters as well as all the runtimes currently do not use the substrate submodule and require a manual synchronization on update.

With this PR dependabot will now update the kagome and gossamer submodule on a weekly basis.

Furthermore, this PR retires all legacy tests, adapters and runtimes and moves the integration test for kagome and gossamer to the non-legacy version.

The work on this caused the following upstream patches:
 - ChainSafe/gossamer#1274
 - ChainSafe/gossamer#1276
 - soramitsu/kagome#651

**Note:** The hostapi-runtime still has to be compiled with an older rust version, because of an unresolved upstream issue caused by wasm blobs compiled with a newer rust version. This issue is tracked in #351.


lamafab (approved):
Went through the PR with Florian via Hangouts. LGTM.

PR #340, opened 2021-01-11
chain-candidates: delete and redirect to official repo

FlorianFranzen:
Instead of storing and updating geneses and runtimes ourself, this PR deletes the outdated files and redirects visitors to the official repository instead.

 This removes the need for any automatic or manual update as suggested in #4.


drskalman:
The idea was that an implementor should be able to get all they need from the Spec repo including the runtime that they need to test their host with. I guess the runtimes are as easily accessible as they were kept in the same repo. So that is OK and the README is clear enough on guiding them to the correct address.

It was cool that we had a wat files but I guess they are not essentials and wasm2wat tools are easily accessible.

drskalman (approved):

PR #342, opened 2021-01-11
gossamer-adapter: implement child storage set/get test

FlorianFranzen:
This implements the most basic child storage test for `ext_default_child_storage_set_version_1` and `ext_default_child_storage_get_version_1`. 

However it fails at the moment because gossamer seems to expect that each child key that is to be used was initialized beforehand with `storage.SetChild(child_key, trie.NewEmptyTrie())` before use, which currently does not happen automatically.

```shell
┌ Error: Adapter failed running `gossamer-adapter host-api --function ext_default_child_storage_get_version_1 --input :child_storage:default:policy,:child_storage:default:full-range,Optional,secondary`:
│ EROR[01-11|19:21:09] [ext_default_child_storage_get_version_1] failed to get child from child storage pkg=runtime module=go-wasmer error="child trie does not exist at key :child_storage:default::child_storage:default:policy" caller=imports.go:498
│ Execution failed:  Failed to call the `rtm_ext_default_child_storage_get_version_1` exported function.
└ @ Main.SpecificationTestsuite.AdapterFixture ~/work/polkadot-spec/polkadot-spec/test/helpers/AdapterFixture.jl:157
```

### DIFF
--- a/test/adapters/gossamer/host_api/child_storage.go
+++ b/test/adapters/gossamer/host_api/child_storage.go
@@ -1,0 +1,130 @@
+// Copyright (c) 2019 Web3 Technologies Foundation
+
+// This file is part of Polkadot Host Test Suite
+
+// Polkadot Host Test Suite is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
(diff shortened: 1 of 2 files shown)


FlorianFranzen:
Disabled for now, just like the child storage tests in kagome. Will be revisited once child storage is properly implemented in both.

lamafab (approved):

PR #347, opened 2021-01-13
ci: upload runtime-spec diff to seperate artifact

FlorianFranzen:
This fixes a "copy-and-waste" error in the CI config, which should result in runtime spec and diff being provided in separate artifacts, as initially intended.


FlorianFranzen:
Works as intended, will merge as the fix is trivial.

PR #352, opened 2021-01-17
ci: only run on push events on master

FlorianFranzen:
Currently we run the CI for pushes to any branch. That means if we open a pull-request from one of our branches to master (which is our default workflow), the CI is run twice, once for the PR and ones for the push to the associated branch.

This PR changes that, by only running for pushes to the master branch. It also removes the filter list for pushes, which means that all workflows always run for pushes to master, which I think makes sense to detect any potential issues. 

If you find it useful to have the Ci run for some of the branches you work on, we could also add wildcards for that, e.g. `staging/**` or similar. This would then mean that you would have to rename the branch before you open a PR to avoid running the CI twice. So not really a perfect solution. It seems however that GitHub can not detect that a push is part of an open PR at the moment.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -2,16 +2,8 @@ name: Integration Testsuite
 
 on:
   push:
-    paths:
-    - '.github/workflows/integration.yml'
-    - 'test/**'
-    - '!test/adapters/**'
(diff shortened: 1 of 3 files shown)


lamafab (approved):

drskalman (approved):
Makes sense

PR #363, opened 2021-01-21
Overhaul Runtime API section

### LINKED ISSUE #328: Complete documentation of Polkadot runtime APIs
Extend the spec with the latest Polkadot runtime APIs.

lamafab:
Fixes #328 .

There are still some open points that need to be added: 

- ~~The Grandpa equivocation report format.~~ tracked in  https://github.com/w3f/polkadot-spec/issues/369
- ~~Spec of Babe equivocations (which is not mentioned at all right now).~~ tracked in  https://github.com/w3f/polkadot-spec/issues/369
- ~~ParachainHost APIs. Currently, the Polkadot master branch just returns empty values. This will probably be submitted in a separate PR.~~ tracked in https://github.com/w3f/polkadot-spec/issues/371

**Printscreen**
ag-runtimeentries

### DIFF
--- a/host-spec/ag-runtimeentries.tm
+++ b/host-spec/ag-runtimeentries.tm
@@ -932,7 +932,7 @@
   <\itemize-dot>
     <item>A SCALE encoded <verbatim|Option> as defined in Definition
     <reference|defn-option-type> containing an array of varying size
-    representing the seed.
+    indicating the seed.
   </itemize-dot>
 
   \;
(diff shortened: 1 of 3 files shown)


FlorianFranzen (commented):
Looks pretty good already, all todos are markt as such.
 
Most of these might be beyond of the scope of this PR, but here are some note:
  - We probably should mention potential panics, as this is apparently how the runtime communicates some failures.
  - Calling `Core_version` should be necessary and we should mention any required checks.
  - `Metadata_metadata` should maybe mention a use case or something like that.
  - `BlockBuilder_apply_extrinsic`
    - Should mention that what it validates i.e. the transaction, not only what it does not.
    - Should mention what to do with each extrinsic based on the returned value
    - E.g. DispatchErrors are still included
    - E.g. InvalidTransactions that exhaust resources might be included next block (?)    
  - `BlockBuilder_check_inherent` and `BlockBuilder_random_seed` lack an explanation on when they should be called.
  -  `TransactionPaymentAPI` looks like you accidentally specced the general call structure again (i.e. scale encoded value + length).

Let me know when the rest is done and I give another quick look.

FlorianFranzen (approved):
Thank you for incorporating all these changes and the mention of the JSON-RPC.

PR #368, opened 2021-01-26
Add-Seal in algorithm 6.7

### LINKED ISSUE #216: Spec should mention the Seal log item addition/removal
In Substrate's code, when we import a block, we first pass it to Aura or Babe.
Aura/Babe look in the block's header for a `Seal` log item, and if one is found they remove it before calling `Core_execute_block`. Then it has to be put back before inserting the block in the database.

Similarly, when producing a block, the runtime does not generate a `Seal` log item. Aura/Babe should do a post-generation-pass to add their seal.

cc @andresilva

lamafab:
Fixes #216.

image

### DIFF
--- a/host-spec/c06-consensus.tm
+++ b/host-spec/c06-consensus.tm
@@ -2,7 +2,7 @@
 
 <project|host-spec.tm>
 
-<style|<tuple|book|algorithmacs-style|old-dots|old-lengths|/home/anon/.TeXmacs/packages/algorithmacs-style.ts>>
+<style|<tuple|book|algorithmacs-style|old-dots|old-lengths>>
 
 <\body>
   <chapter|Consensus><label|chap-consensu>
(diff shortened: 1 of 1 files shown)


host-spec/c06-consensus.tm
```diff
@@ -1,8 +1,8 @@
-<TeXmacs|1.99.17>
+<TeXmacs|1.99.16>
 
 <project|host-spec.tm>
 
-<style|<tuple|book|algorithmacs-style|old-dots|old-lengths>>
+<style|<tuple|book|algorithmacs-style|old-dots|old-lengths|/home/anon/.TeXmacs/packages/algorithmacs-style.ts>>
```
FlorianFranzen: This probably should be reverted.

FlorianFranzen (approved):
Please just remove the reference to your home directory and it should be fine to merge.

host-spec/c06-consensus.tm
lamafab: Done!

PR #378, opened 2021-02-02
Clearly indicate when transactions should be rejected.

### LINKED ISSUE #367: Clearly indicate when transactions should be rejected.
In the specification of the `BlockBuilder_apply_extrinsics`, it should clearly indicate based on the return value which transactions should be accepted, rejected or queued for later.

lamafab:
Fixes #367.

## Printscreens

image

And adjusted Algorithm 6.7.

image


FlorianFranzen (approved):
Very nice, this should make it very clear how the host should handle failures.

PR #379, opened 2021-02-02
Short definition of Dec_SC

### LINKED ISSUE #58: Define Dec-SC parallel to Enc-SC
We need to just mention that Dec_SC is the decoding for Scale codec and just revert Enc_SC

lamafab:
Fixes #58.

There really isn't much else to say. We also don't use it anywhere, as of now.

## Printscreen

image


FlorianFranzen (approved):
Thanks, that should be good enough for now.

Issue #161, opened 2020-04-06
Define last completed round in contrast to last finalized round


drskalman:
In GRANDPA and specifically in the code there is three different concepts of rounds:

 - completable round
 - completed round 
 - finalized round 

We need to make the difference of these three concepts clear in the spec.

drskalman:
1. completable round is the same as the spec definition of it in Definition 6.3.
2. We don't have notation of completed round, the node just need to complete algorithm 6.9 which basically means the round is completed.
3. We don't call a round to be finalized neither they code does, we both only call a block to be finalized.

FlorianFranzen:
@drskalman There is still a todo for this in section 6.4.1.2 (Processing catch-up requests) of the host-spec in place of a definition of "Last-Completed-Round", Could you please take a look at this and see what needs to be done?

drskalman:
Checked @FlorianFranzen's comment. replaced the todo and updated relevant algorithms: 8cc23e11. Needs @AlistairStewart review.

drskalman:
I sought @AlistairStewart oral review. Everything is correct. Except, rejoining should happen in the same round that the node has left. I'll update the spec and the pull request.

drskalman:
Following my talk to @AlistairStewart, if you disconnect and connect again then you should keep with the last round
    you had before nothing changes. If you are starting a fresh authority set then last round is 0.

    This is reflected in 44a6b26c

    The rest of changes was approved verbally.

PR #380, opened 2021-02-03
Define non-voter node and clarify justification for it vs voter

### LINKED ISSUE #133: Make the diffinition of justification more accurate
Justification is defined in grandpa spec and is basically a vector of votes. The same structure  is used for pre-vote and pre-commit. To comply with recent GRANDPA development, we need to:
  - [x] adapt the justification and whereever it is used in the text to to have both stages (pre-vote and pre-commit). rformerly it was used only for pre-commit.
  - [x] Make sure the format accurately reflect the format currently used in GRANPDA.
  - [x] Make the distinction between the base of a justification and the highest block that it finlized. Now there is a definition of a base block (which is the pre-vote ghost of last round) and that shouldn't be confused by the block that a justification finalizes. As GRANDPA finalises subgraphs and not blocks.

drskalman:
Fixes #133.

### DIFF
--- a/host-spec/c06-consensus.tm
+++ b/host-spec/c06-consensus.tm
@@ -1,4 +1,4 @@
-<TeXmacs|1.99.16>
+<TeXmacs|1.99.17>
 
 <project|host-spec.tm>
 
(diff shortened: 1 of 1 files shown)


FlorianFranzen (approved):
Looks good. LGTM!

Does this also take care of #161?

drskalman:
Verified #161 and it appears the Spec is good to go in current form so I closed #161.

PR #397, opened 2021-02-24
hostapi-runtime: Fix double scale encoding

FlorianFranzen:
This removes the last remaining double encoded values returned by the `hostapi-runtime`. For consistency it also changes any dynamically sized inputs to their correct fixed sizes.

This subsequently allowed us to removes all the double encoding fixes from `substrate-adapter` and `kagome-adapter`.


lamafab (approved):

lamafab (commented):
:+1:

PR #402, opened 2021-03-03
host-spec: Various cleanups

FlorianFranzen:
This is a collection of various smaller changes to the host spec:
 - remove leftover version diff tag
 - delete legacy hostapi
 - fixup appendix lettering
 - add generated figure to gitignore
- rename runtime entries to api
- fix broken references
- fix various mistakes in host api
- update inherents section
- document `__heap_base`
- document `:heappages`;
- document the use of imported memory
- extend runtime API with version info
- restructure runtime api sections
- clean up tables in runtime api chapter
- remove all indices from chapters


drskalman:
- Collection vs Byte array I didn't understand what was going on
 - "rename runtime entries to api
" I remember we spceifically discussed not to do that as these are function called by the host where there many other runtime apis which are called within the  runtime. We said we'll find a approperiate way of distinguishing the two in the startegy meeting.
 - Pushed here because I don't if I messed up with the labels or not
 https://github.com/w3f/polkadot-spec/commit/e66f2155ab74ffbb4fae833549cf66472568ae51

FlorianFranzen:
I will fix the typo and defined SCALE collection.

lamafab:
Went over this with Florian in a Google Meet call. LGTM.

PR #408, opened 2021-03-08
testsuite: Test more then just default wasm backend

FlorianFranzen:
This extends the testsuite to also allow the test of the none-default backends. This is achieved by using the optional `environment` flag the substrate and gossamer adapter now support.

While working on the `gossamer-adapter` I also added a `runtime` flag to the to simplify debugging, retired our custom error types to us the go 13 error wrapping and added support for the new life backend.


FlorianFranzen:
@lamafab approves.

Issue #404, opened 2021-03-03
Pre-Publication Review of Host Spec


FlorianFranzen:
# Skipped chapters:

- C07 AnV
- C08 Messaging

# Stage 1 - Spec Team

- [x] C01 Background @lamafab (#419)
- [x] C02 State @lamafab  (#419)
- [x] C03 Transition @FlorianFranzen (#439)
- [x] C04 Networking @drskalman (#435)
- [x] C05 Bootstrapping @FlorianFranzen
- [x] C06 Consensus  @drskalman  (#435)
- [x] A Cryptoalgorithms @drskalman (#446)
- [x] B Encodings @FlorianFranzen 
- [x] C Genesis @lamafab (#419)
- [x] D Hostapi @FlorianFranzen 
- [x] E Runtimeapi @drskalman (#446)

# Stage 2 - Research Team

- [x] C01 Background  @AlfonsoCev 
- [x] C02 State @AlfonsoCev 
- [ ] C03 Transition @AlfonsoCev 
- [x] C04 Networking @infinity0 -> #403
- [x] C06 Consensus @hndnklnc (through pdf/florian) @AlistairStewart (through Syed)

drskalman:
- [ ] C04 network @skalman's review -> f3ee88da

FlorianFranzen:
>     * [ ]  C04 network @Skalman's review -> [f3ee88d](https://github.com/w3f/polkadot-spec/commit/f3ee88da4868a9e3117f377adb409b02a8f31321)

This was included in #435. I updated the issue accordingly.

drskalman:
Encoding is almost done. We might change the ordering based on @AlfonsoCev 's feedback.

FlorianFranzen:
Apart from chapter 3, all chapters underwent some basic review.

PR #419, opened 2021-03-16
Chapter review (assigned to Fabio)

lamafab:
Review/adjustment of my assigned chapters from #404.

* C01 Background @lamafab
* C02 State @lamafab
* C Genesis @lamafab

I do have some open questions, which I did not write directly into the specification. I think it's better to have those listed here, which probably must be answered directly by @drskalman.

***
* Section 1.2, Definition 1.1 (page 13): Discrete State Machine (DSM) is defined once and never used again. Is this necessary?
* Section 1.2, Definition 1.4 (page 14): Do we really need to define "sequence of bytes"?
* Section 1.2, Definition 1.6 (page 14): b (bitwise representation) = b (sequence of of bytes, Definition 1.4)? -> variable name collision.
* Section 1.2, Definition 1.7 (page 14): B = b (Definition 1.4)
* Section 2.1.2 (page 17): "_When traversing the Trie to a specific node, its key can be reconstructed by concatenating the subsequences of the key which are stored either explicitly in the nodes on the path or implicitly in their position as a child of their parent."_

  Clarify this?
* ~~Section 2.1.4 (page 19): How exactly are the values in storage ordered (during computation, not in storage) to compute the final merkle proof in a reproducible way?~~

### DIFF
--- a/host-spec/ac-genesis.tm
+++ b/host-spec/ac-genesis.tm
@@ -1,25 +1,26 @@
-<TeXmacs|1.99.12>
+<TeXmacs|1.99.16>
 
 <project|host-spec.tm>
 
-<style|book>
+<style|<tuple|book|old-lengths>>
 
 <\body>
   <appendix|Genesis State Specification><label|sect-genesis-block>
 
   The genesis state represents the intial state of Polkadot state storage as
   a set of key-value pairs, which can be retrieved from
-  <cite|web3.0_technologies_foundation_polkadot_2020>. While each of those
-  key/value pairs offer important identifyable information which can be used
-  by the Runtime, from the Polkadot Host points of view, it is a set of
-  arbitrary key-value pair data as it is chain and network dependent.
-  \ Except for the <verbatim|:code> described in Section
-  <reference|sect-loading-runtime-code> which needs to be identified by the
-  Polkadot Host to load its content as the Runtime. The other keys and values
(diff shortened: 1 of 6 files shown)


FlorianFranzen:
>     * Section 1.2, Definition 1.1 (page 13): Discrete State Machine (DSM) is defined once and never used again. Is this necessary?

This is fine and just used to embed the spec into the theoretical informatics literature.

>     * Section 1.2, Definition 1.4 (page 14): Do we really need to define "sequence of bytes"? 
>     * Section 1.2, Definition 1.6 (page 14): b (bitwise representation) = b (sequence of of bytes, Definition 1.4)? -> variable name collision.
>     * Section 1.2, Definition 1.7 (page 14): B = b (Definition 1.4)

These are a bit basic, we might want to move them in an appendix. Lets discuss this and the variable collision in the strategy meeting.
 
>     * Section 2.1.2 (page 17): "_When traversing the Trie to a specific node, its key can be reconstructed by concatenating the subsequences of the key which are stored either explicitly in the nodes on the path or implicitly in their position as a child of their parent."_

Indeed this is a bit convoluted, but it is technically correct. I would leave it for now. 

>     * ~Section 2.1.4 (page 19): How exactly are the values in storage ordered (during computation, not in storage) to compute the final merkle proof in a reproducible way?~

I think the order is just the same and therefore it should be clear from the previous definitions.

PR #433, opened 2021-03-31
[WIP] Spec beefy grandpa

### LINKED ISSUE #331: Spec part of beefy protocol which is executed by Polkadot host validator node
References for this can be found:
 1. https://github.com/paritytech/grandpa-bridge-gadget
 2. https://hackmd.io/ohOt4jAPT8uu-soJXHUq0Q

lamafab:
Fixes #331.

As mentioned in our sync meeting, the two entities "Relayer" and "Light client" are quite abstract. Would appreciate it you, @FlorianFranzen , could tell me whether the current version makes sense to you, even if broad. I will also ask Alistair on what he thinks. I also believe BEEFY would appreciate my charting-skill :)

### Chapter 4: Networking

image
image

### Chapter 6: Consensus

image
image

### DIFF
--- a/host-spec/c04-networking.tm
+++ b/host-spec/c04-networking.tm
@@ -736,35 +736,83 @@
   </todo>
 
   <\definition>
-    A signed commitment is a datastructure of the following format:
+    <label|defn-grandpa-beefy-commitment>A commitment, <math|C>, contains the
+    information extracted from the finalized block at height <math|B<rsub|n>>
+    as specified in the message body. <todo|explain some more>
+
+    C is a datastructe of the following format:
 
     <\eqnarray*>
-      <tformat|<table|<row|<cell|M<rsub|BFY>>|<cell|=>|<cell|Enc<rsub|SC><around*|(|C,S<rsub|n>|)>>>|<row|<cell|C>|<cell|=>|<cell|<around*|(|M<rsub|mr>,B<rsub|n>,id<rsub|\<bbb-V\>>|)>>>|<row|<cell|S<rsub|n>>|<cell|=>|<cell|<around*|(|V<rsup|sig><rsub|0>,\<ldots\>,V<rsup|sig><rsub|n>|)>>>>>
+      <tformat|<table|<row|<cell|C>|<cell|=>|<cell|<around*|(|R,B<rsub|n>,id<rsub|\<bbb-V\>>|)>>>>>
     </eqnarray*>
+
+    where
+
+    <\itemize-dot>
+      <item><math|R> is the MMR root of all the block header hashes leading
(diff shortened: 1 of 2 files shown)


FlorianFranzen:
~~Will need to be rebased against new and filtered main branch.~~

FlorianFranzen:
Here are some notes:
 - I probably would us `H_i(B_{last})` instead of `B_n`.
 - I would prefix more definitions with "BEEFY", e.g. "BEEFY vote message", "BEEFY light client"
 - If a variable contains multiple items, `Var_{0...n}` instead of `Var_n` might make this more clear
 - Lets hope the `S_n` data structure is revised...
   - Might be worth to put in its own definition, e.g. as Signature Set, etc.  
   - "as specified by id_v in C" -> "(Definition <defn-grandpa-...>)"
 - Is `V_n` encoded as a bitfield?
 - The preliminary section should probably contain more symbols (`R_h`, `R_{sig}`, etc.)
 - I would probably remove any mentions of Ethereum.
 - You mention witness data on chain, but later say this is undefined. 
 - In general, relayers and beefy light clients are not really relevant to the host spec and I would just move them to a bridge spec. From your write-up, all the host needs to do is gossip BEEFY votes. The aggregation and distribution is external to the host and will depend heavily on the network it connects to.

lamafab:
> Lets hope the `S_n` data structure is revised... 

I'm leaving it as is, but referenced the definition of `id_v`. Since the signature set (`S_n`) is currently just used here, I don't think it needs its own definition as of now.

> Is `V_n` encoded as a bitfield?

Not as of now:

```rust
pub signed_by: Vec<bool>, // TODO [ToDr] Consider replacing with bitvec crate
```

> The preliminary section should probably contain more symbols (R_h, R_{sig}, etc.)

Why? Those are defined in the message definition.

> I would probably remove any mentions of Ethereum.

I think mentioning Ethereum as an example can demonstrate how the commitments/witnessdata can be saved on chain in a non-trusted way. So I'd leave it in there.

> You mention witness data on chain, but later say this is undefined.

Where? The *bridge* is undefined, meaning there's no clear definition on *how* the commitments/witnessdata are *sent* to the remote network.

> In general, relayers and beefy light clients are not really relevant to the host spec and I would just move them to a bridge spec. From your write-up, all the host needs to do is gossip BEEFY votes. The aggregation and distribution is external to the host and will depend heavily on the network it connects to.

That's true, but I think here in the Host spec it should at least explain the context. Or maybe we can just specify the network messages, and do the rest in a bridge spec, not sure. My proposal: let's currently keep those concepts in the Host spec and once BEEFY/bridges become more mature we can put it into a bridge spec, if we think that's appropriate then.

PR #438, opened 2021-04-07
host-spec: update grandpa messages

### LINKED ISSUE #131: Spec Neighbor packet message
It is the packet that trigger grandpa catchup request beside other reasons that it needs to be specced.

Implementation:
  - [Parity Polkadot](https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/communication/gossip.rs#L311)

```rust
/// V1 neighbor packet. Neighbor packets are sent from nodes to their peers
/// and are not repropagated. These contain information about the node's state.

#[derive(Debug, Encode, Decode, Clone)]
pub(super) struct NeighborPacket<N> {
	/// The round the node is currently at.
	pub(super) round: Round,
	/// The set ID the node is currently at.
	pub(super) set_id: SetId,
	/// The highest finalizing commit observed.
	pub(super) commit_finalized_height: N,
}
```

FlorianFranzen:
After I recently moved the Grandpa messages into the network chapter, I stumbled across a few issue, which this PR should fix:

- Pre-commit incorrectly use commit message type id
- Finalization messages is not identified as commit message
- Finalization/Commit message does not use compact justification format

Other changes:

- Add Neighbor message definition (fixes #131)
- Move some of the definitions back into chapter 6
- Specify which messages get re-gossiped.
- FIx bootstrap chapter references

### DIFF
--- a/host-spec/c04-networking.tm
+++ b/host-spec/c04-networking.tm
@@ -301,7 +301,7 @@
     protocol that sends GRANDPA votes to connected peers. This is a
     <em|Notification substream>.
 
-    The messages are specified in Section <reference|sect-msg-grandpa-vote>.
+    The messages are specified in Section <reference|sect-msg-grandpa>.
 
     <todo|This substream will change in the future. See <hlink|issue
     #7252|https://github.com/paritytech/substrate/issues/7252>.>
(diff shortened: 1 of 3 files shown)


lamafab:
:+1: 

Do note, however, that there are some issues with the width of some tables.

image
image

FlorianFranzen:
Could not reproduce table width issue, will merge for now.

PR #439, opened 2021-04-07
host-spec: chapter 3 review

FlorianFranzen:
This is my review of chapter 3. 

I also tried to incorporate the fact that offchain workers can submit extrinsics as well a bit better, though it is not perfect yet.

### DIFF
--- a/host-spec/ae-runtimeapi.tm
+++ b/host-spec/ae-runtimeapi.tm
@@ -514,12 +514,15 @@
 
   <subsubsection|<verbatim|TaggedTransactionQueue_validate_transaction>><label|sect-rte-validate-transaction>
 
-  This entry is invoked against extrinsics submitted through the transaction
-  network message <reference|sect-msg-transactions> and indicates if the
-  submitted blob represents a valid extrinsics applied to the specified
-  block. This function gets called internally when executing blocks with the
-  <verbatim|Core_execute_block> runtime function as described in section
-  <reference|sect-rte-core-execute-block>.
+  This entry should be invoked with all extrinsics submitted through a
+  transaction network message <reference|sect-msg-transactions> or by an
+  offchain worker through the <verbatim|ext_offchain_submit_transaction> Host
+  API (Section <reference|sect-ext-offchain-submit-transaction>). It will
+  indicates if the submitted blob represents a valid extrinsics applied to
+  the specified block, how it depends on other extrinsics and if it should be
+  gossiped to other peers. Furthermore this function gets called internally
+  when executing blocks with the <verbatim|Core_execute_block> runtime
+  function as described in section <reference|sect-rte-core-execute-block>.
 
   \;
 
@@ -550,7 +553,7 @@
     </big-table>
   </definition>
 
-  <strong|Return>:This function returns a <verbatim|Result> as defined in
+  <strong|Return>: This function returns a <verbatim|Result> as defined in
   Definition <reference|defn-result-type> which contains the type
   <em|<verbatim|ValidTransaction>> as defined in Definition
   <reference|defn-valid-transaction> on success and the type
@@ -577,7 +580,7 @@
     the validity to be correct>|<cell|>>|<row|<cell|Longevity>|<cell|After
     this period, the transaction should be removed from the >|<cell|Unsigned
     64bit>>|<row|<cell|>|<cell|pool or revalidated.>|<cell|integer>>|<row|<cell|Propagate>|<cell|A
-    flag indicating if the transaction should be propagated to
+    flag indicating if the transaction should be gossiped to
     >|<cell|Boolean>>|<row|<cell|>|<cell|other peers.>|<cell|>>>>>>
       The tuple provided by <verbatim|TaggedTransactionQueue_transaction_validity>
 
@@ -587,7 +590,7 @@
 
   <strong|Note>: If <em|Propagate> is set to <verbatim|false> the transaction
   will still be considered for including in blocks that are authored on the
-  current node, but will never be sent to other peers.
+  current node, but should not be gossiped to other peers.
 
   \;
 
--- a/host-spec/ae-runtimeapi.tm
+++ b/host-spec/ae-runtimeapi.tm
@@ -517,12 +517,13 @@
   This entry should be invoked with all extrinsics submitted through a
   transaction network message <reference|sect-msg-transactions> or by an
   offchain worker through the <verbatim|ext_offchain_submit_transaction> Host
-  API (Section <reference|sect-ext-offchain-submit-transaction>). It will
+  API (Section <reference|sect-ext-offchain-submit-transaction>). It
   indicates if the submitted blob represents a valid extrinsics applied to
-  the specified block, how it depends on other extrinsics and if it should be
-  gossiped to other peers. Furthermore this function gets called internally
-  when executing blocks with the <verbatim|Core_execute_block> runtime
-  function as described in section <reference|sect-rte-core-execute-block>.
+  the specified block, in which order it should be applied and if it should
+  be gossiped to other peers. Furthermore this function gets called
+  internally when executing blocks with the <verbatim|Core_execute_block>
+  runtime function as described in section
+  <reference|sect-rte-core-execute-block>.
 
   \;
 
@@ -589,7 +590,7 @@
   </definition>
 
   <strong|Note>: If <em|Propagate> is set to <verbatim|false> the transaction
-  will still be considered for including in blocks that are authored on the
+  will still be considered for inclusion in blocks that are authored on the
   current node, but should not be gossiped to other peers.
 
   \;
--- a/host-spec/c03-transition.tm
+++ b/host-spec/c03-transition.tm
@@ -7,37 +7,39 @@
 <\body>
   <chapter|State Transition><label|chap-state-transit>
 
-  Like any transaction-based transition system, Polkadot state changes via
-  executing an ordered set of instructions. These instructions are known as
-  <em|extrinsics>. In Polkadot, the execution logic of the state-transition
-  function is encapsulated in Runtime as defined in Definition
-  <reference|defn-state-machine>. Runtime is presented as a Wasm blob in
-  order to be easily upgradable. Nonetheless, the Polkadot Host needs to be
-  in constant interaction with Runtime. The detail of such interaction is
-  further described in Section <reference|sect-entries-into-runtime>.
+  Like any transaction-based transition system, Polkadot's state is changed
+  by executing an ordered set of instructions. These instructions are known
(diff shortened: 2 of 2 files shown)


host-spec/ae-runtimeapi.tm
```diff
@@ -514,12 +514,15 @@
...
-  <reference|sect-rte-core-execute-block>.
+  This entry should be invoked with all extrinsics submitted through a
+  transaction network message <reference|sect-msg-transactions> or by an
+  offchain worker through the <verbatim|ext_offchain_submit_transaction> Host
+  API (Section <reference|sect-ext-offchain-submit-transaction>). It will
+  indicates if the submitted blob represents a valid extrinsics applied to
+  the specified block, how it depends on other extrinsics and if it should be
+  gossiped to other peers. Furthermore this function gets called internally
+  when executing blocks with the <verbatim|Core_execute_block> runtime
+  function as described in section <reference|sect-rte-core-execute-block>.
```
lamafab: > It will indicates ...

should be:

> It will **indicate** ...

But personally I'd write it as:

> It indicates whether ...

Also, I don't understand this part:

> ..., how it depends on other extrinsics ...

Can you clarify what you mean?

host-spec/ae-runtimeapi.tm
```diff
@@ -587,7 +590,7 @@
 
   <strong|Note>: If <em|Propagate> is set to <verbatim|false> the transaction
   will still be considered for including in blocks that are authored on the
```
lamafab: > ... will still be considered for **inclusion** ...

maybe?

host-spec/c03-transition.tm
```diff
@@ -7,37 +7,39 @@
...
-  function is encapsulated in Runtime as defined in Definition
-  <reference|defn-state-machine>. Runtime is presented as a Wasm blob in
-  order to be easily upgradable. Nonetheless, the Polkadot Host needs to be
-  in constant interaction with Runtime. The detail of such interaction is
-  further described in Section <reference|sect-entries-into-runtime>.
+  Like any transaction-based transition system, Polkadot's state is changed
+  by executing an ordered set of instructions. These instructions are known
+  as <em|extrinsics>. In Polkadot, the execution logic of the
+  state-transition function is encapsulated in a Runtime as defined in
+  Definition <reference|defn-state-machine>. For easy upgradability this
```
lamafab, line 14: > For **easier** upgradability **and in order to avoid hard forks in the Polkadot network,** this ...

We'd probably need to define "hard forks", then. Up to you.

host-spec/c03-transition.tm
```diff
@@ -7,37 +7,39 @@
...
   Runtime and queued to be applied to the current state.
 
-  Polkadot, as with most prominent distributed ledger systems that make state
-  replication feasible, journals and batches a series of extrinsics together
-  in a structure known as a <em|block> before propagating to the other nodes.
-  The specification of the Polkadot block as well as the process of verifying
-  its validity are both explained in Section
+  \ To make state replication feasible, Polkadot journals and batches series
+  of its extrinsics together into a structure known as a <em|block>, before
+  propagating them to other nodes, similar to most other prominent
```
lamafab, line 26: In previous changes you explicitly used "gossip", now you use "propagate" ¯\\\_(ツ)\_/¯

host-spec/c03-transition.tm
```diff
@@ -81,37 +83,32 @@
...
-  Subsequent calls to the runtime have the ability to, in turn, call the
-  storage API (see Section <reference|sect-host-api>) to insert a new Wasm
-  blob
-
-  into runtime storage slot to upgrade the runtime.
+  The initial runtime code of the chain is provided as part of the genesis
+  state (see Section <reference|sect-genesis-block>) and subsequent calls to
+  the Runtime have the ability to, in turn, call the storage API (see Section
+  <reference|sect-host-api>) to insert new Wasm blobs into runtime storage to
+  upgrade the Runtime.
```
lamafab: > **R**untime

missed it twice.

> ... to insert new Wasm blobs into runtime storage to upgrade the Runtime.

Maybe:

>  ... to insert a new, upgraded Runtime Wasm blob.

?

host-spec/c03-transition.tm
```diff
@@ -157,18 +146,18 @@
...
   the Polkadot Runtime.
 
-  The size of the provided WASM memory should be based on the
-  <verbatim|:heappages> storage key, where each page has the size of 64KB.
-  This memory shoule be made available to the Polkadot runtime for import
-  under the symbol name <verbatim|memory>.
+  The size of the provided WASM memory should be based on the value of the
+  <verbatim|:heappages> storage key (an unsigned 64-bit integer), where each
+  page has the size of 64KB. This memory shoule be made available to the
+  Polkadot runtime for import under the symbol name <verbatim|memory>.
```
lamafab: > **R**untime

host-spec/c03-transition.tm
```diff
@@ -157,18 +146,18 @@
...
 
-  <subsubsection|Sending Arguments to Runtime
+  <subsubsection|Sending Data to a Runtime Entry
   ><label|sect-runtime-send-args-to-runtime-enteries>
 
   In general, all data exchanged between the Polkadot Host and the Runtime is
   encoded using SCALE codec described in Section
-  <reference|sect-scale-codec>. As a Wasm function, all runtime entries have
-  the following identical signatures:
+  <reference|sect-scale-codec>. Therefore all runtime entries have the
```
lamafab, line 159: > Therefore, all

Might also be useful to say that communication between the Runtime and Host API tries to avoid SCALE as much as possible in order to increase performance.

host-spec/c03-transition.tm
```diff
@@ -258,34 +244,31 @@
...
+
+  Alternative transaction can be submitted to the host by offchain worker
+  through the <verbatim|ext_offchain_submit_transaction> Host API, defined in
+  Section <reference|sect-ext-offchain-submit-transaction>.\ 
+
+  Any new transaction should be submitted to the
+  <verbatim|validate_transaction> Runtime function, defined in Section
+  <reference|sect-rte-validate-transaction>. This will allow the Polkadot
+  Host to check the validity of the received transaction against the current
+  state as well as determine how the transaction depends on other extrinsics
```
lamafab: > the transaction depends on other extrinsics

again, I don't understand this sentence :thinking:

host-spec/c03-transition.tm
```diff
@@ -626,26 +613,26 @@
 
   <subsection|Importing and Validating Block><label|sect-block-validation><label|sect-block-submission>
 
-  Block validation is the process by which the client asserts that a block is
-  fit to be added to the blockchain. This means that the block is consistent
-  with the world state and transitions from the state of the system to a new
-  valid state.
+  Block validation is the process by which a node asserts that a block is fit
+  to be added to the blockchain. This means that the block is consistent with
+  the world state and transitions from the current state of the system to a
```
lamafab: > world state

I know what you mean but this is an *interesting* term to use here. Might require clarification. Maybe just:

> This means that the block is consistent with the **current state and can transition to a new valid state**.

host-spec/c03-transition.tm
```diff
@@ -626,26 +613,26 @@
...
+  Runtime and the Polkadot Host then need to work together to assure block
+  validity. A block is deemed valid if the block author had authorship rights
+  for the slot in which the block was produce as well as if the transactions
+  in the block constitute a valid transition of states. The former criterion
+  is validated by the Polkadot Host according to the block production
+  consensus protocol. The latter can be verified by the Polkadot Host
+  invoking <verbatim|Core_execute_block> entry into the Runtime as defined in
+  section <reference|sect-rte-core-execute-block> as a part of the validation
+  process. Any state changes created by this function on successful execution
+  are persisted.
```
lamafab, line 633: > persisted

Basti hates this word :) Basically, executing a block simply creates a new state (state root), which the state machine can navigate to/from. I guess this is fine as it is. Let me know.

host-spec/c03-transition.tm
```diff
@@ -831,9 +815,9 @@
     V<rsub|C>=Enc<rsub|SC><around*|(|C<rsub|value>|)>
   </equation*>
 
-  is SCALE encoded byte array.
+  is a SCALE encoded byte array.
 
-  where <math|K> is the changed storage key,
+  Furthermore <math|K> represents the changed storage key,
```
lamafab: > Furthermore, 

(comma)

lamafab (changes requested):

host-spec/c03-transition.tm
FlorianFranzen: Good to know. But as far as I can tell, "persisted" was also used here before.

host-spec/c03-transition.tm
FlorianFranzen: It used the term "world state" also before my changes, but I agree that it probably should be changed, as we do not use this term anywhere else.

host-spec/c03-transition.tm
lamafab: Yep, I think I originally wrote that.

host-spec/c03-transition.tm
lamafab: You can blame this one on Syed.

host-spec/c03-transition.tm
FlorianFranzen: I would avoid that for now.

host-spec/c03-transition.tm
FlorianFranzen: I tried no to repeat myself to much.

host-spec/c03-transition.tm
FlorianFranzen: Thanks, Fixed.

host-spec/c03-transition.tm
FlorianFranzen: That might be true for the Host API, but given that the Runtime API encodes EVERYTHING passed to and returned from a function, I do no think it is universally true.

host-spec/c03-transition.tm
FlorianFranzen: `validate_transaction` returns this information, so I thought it should be mentioned. I will remove it for now.

host-spec/c03-transition.tm
FlorianFranzen: Fixed.

host-spec/ae-runtimeapi.tm
FlorianFranzen: I replaced it with "in which order it should be applied" to make it more clear what I mean with "depend on"

host-spec/ae-runtimeapi.tm
FlorianFranzen: Yes, for sure!

PR #447, opened 2021-04-13
Skalman glossary integration

### LINKED ISSUE #105: Convert glossary to latex and then import into texmacs
The glossary is in mark down and can be found here right now here
   https://github.com/lamafab/polkadot-spec-glossary

drskalman:
Close #105 . Work in progress. Chapter 01 is done.


drskalman:
It should be done. Please review.

FlorianFranzen:
Manually went through all commits to fix various changes caused by initial version being based on host spec before chapter review. Will merge after one last review of diff.

FlorianFranzen (approved):
I manually went through all changes in multiple rounds of review and cleanups. LGTM!

PR #199, opened 2020-05-20
Import overall UML, split into three parts

lamafab:
I had to split the UML into three parts to make it work. Would also like to know what @AlistairStewart thinks about this.

I'd consider having this in separate files (as a raw PlantUML file and a PNG).

[...]

PR #449, opened 2021-04-27
host-spec: split and extend AnV flowcharts

FlorianFranzen:
This PR recovers the UML files from #199 and adopts them to the new file structure, so we can decide if we still want to include them in the spec.


FlorianFranzen:
Closed, no longer relevant.

PR #453, opened 2021-05-11
Fix bool types

lamafab:


FlorianFranzen (approved):
Very nice catch. Having the Rust types translated by a macro does not make things easier.

PR #459, opened 2021-07-06
host-spec: add html release

FlorianFranzen:
This PR add HTML to the list of the supported export formats of the host-spec.

It currently lack CI support, I will add that in the next few days.


FlorianFranzen:
Latest release was successful. LGTM!

PR #460, opened 2021-07-24
Fix return type of ext_storage_root and ext_storage_changes_root

Harrm:
As far as I can see, the runtime expects these methods to return i64. 
https://github.com/paritytech/substrate/blob/master/primitives/externalities/src/lib.rs#L158
In addition, changes root is optional.
https://github.com/paritytech/substrate/blob/master/primitives/externalities/src/lib.rs#L178


FlorianFranzen:
@Harrm Thank you for your contribution. We will fix this in the new spec as well.

PR #465, opened 2021-08-18
Skalman adjust anv ec review

drskalman:
Because @lamafab said so...


drskalman:
Looks good to @lamafab

PR #473, opened 2021-12-15
Convert all internal references to xref references

lamafab:
As discussed in the Spec Team channel.


lamafab:
As discussed in the meeting, this is being retired for now.

PR #478, opened 2022-02-21
New trie version

lamafab:
Adds:

* New trie node header format.
* Note on how to calculate the Merkle proof with "node hashes".
* `state_version` field in `Core_version` runtime function.
* `ext_storage_root_version_2` Host API function.


lamafab:
Note that this must be extended the the `child_storage` and `trie` too.

FlorianFranzen (approved):
I will probably split the previous and the new extension header before merging into master, to be able to point from it from the changelog.

PR #487, opened 2022-03-22
Reference Core_version from Host API.

lamafab:


FlorianFranzen (approved):
Just call "some" function, it'll be fine.

PR #490, opened 2022-03-23
Skalman fix grandpa algorithms

drskalman:


FlorianFranzen:
Integrated the algorithms with new pseudocode renderer, will review rest later.

FlorianFranzen (approved):
Squashed the PDF out of the git history. Otherwise looks good.

PR #491, opened 2022-03-24
Define secondary slots

### LINKED ISSUE #233: Specify Babe's secondary slots
There are currently [two secondary slot modes](https://github.com/paritytech/substrate/blob/18334ee173f0ec4b62d2ca05c5b0c1f0b05b1b17/primitives/consensus/babe/src/lib.rs#L218-L227) in substrate and both are in use (plain in Kusama, VRF in Polkadot).

lamafab:
Fixes #233.

Note that I'll have to update the primary slots section since we expanded our VRF definitions since it was last written. I will do that in a separate PR.

image

image

image

### DIFF
--- a/01_host/05_consensus/block_production.adoc
+++ b/01_host/05_consensus/block_production.adoc
@@ -213,7 +213,8 @@ p larr "dleq_prove"(t_6, h)
 ++++
 
 The operators are defined in <<defn-strobe-operations>>, stem:["dleq_prove"] in
-<<defn-vrf-dleq-prove>>.
+<<defn-vrf-dleq-prove>>. The computed outputs, stem:[o] and stem:[p], are
+included in the block Pre-Digest (<<defn-babe-header>>).
 ****
 
 [#sect-slot-number-calculation]
--- a/01_host/05_consensus/block_production.adoc
+++ b/01_host/05_consensus/block_production.adoc
@@ -177,7 +177,9 @@ where:
 * stem:[h(...)] creates a 256-bit Blake2 hash from its inner value.
 * stem:[A_l] is the lengths of the authority list (<<defn-authority-list>>).
 
-To create the necessary VRF outputs, a transcript is construced (<<sect-vrf>>).
+If stem:[i_d] points to the authority, that authority must claim the secondary
+slot by calculataing the necessary outputs for block production. To create the
+VRF outputs, a transcript is construced (<<sect-vrf>>).
 
 [stem]
 ++++
--- a/01_host/05_consensus/block_production.adoc
+++ b/01_host/05_consensus/block_production.adoc
@@ -170,6 +170,25 @@ where:
 * stem:[h(...)] creates a 256-bit Blake2 hash from its inner value.
 * stem:[A_l] is the lengths of the authority list (<<defn-authority-list>>).
 
+Then a transcript...
+
+[stem]
+++++
+t_1 larr "Transcript"("'BABE'")\
+t_2 larr "append"(t_1, "'slot number'", s)\
+t_3 larr "append"(t_2, "'current epoch'", e_i)\
+t_4 larr "append"(t_3, "'chain randomness'", r)\
+t_5 larr "append"(t_4, "'vrf-nm-pk'", p_k)\
+t_6 larr "meta-ad"(t_5, "'VRFHash'", "False")\
+t_7 larr "meta-ad"(t_6, 64_("le"), "True")\
+i larr "prf"(t_7, "False")\
+o = s_k * i\
+p larr sf "dleq_prove"(t_6, i)
+++++
+
+The resulting value stem:[p] is the VRF proof (<<defn-vrf-proof>>).
+stem:["dleq_prove"] is described in <<defn-vrf-dleq-prove>>.
+
 ****
 
 [#sect-slot-number-calculation]
--- a/01_host/05_consensus/block_production.adoc
+++ b/01_host/05_consensus/block_production.adoc
@@ -98,7 +98,8 @@ the lottery.
 In order to ensure consistent block production, BABE uses secondary slots in
 case no authority won the (primary) block production lottery. Unlike the
 lottery, secondary slot assignees are know upfront publically
-(<<block-production-secondary-slots>>).
+(<<block-production-secondary-slots>>). The Runtime dictates what types of
+rounds to execute (<<sect-rte-babeapi-epoch>>).
 
 [#defn-babe-constant]
 .<<defn-babe-constant, BABE Constant>>
--- a/01_host/05_consensus/block_production.adoc
+++ b/01_host/05_consensus/block_production.adoc
@@ -141,14 +141,18 @@ the lottery.
 Algorithm: stem:["Block-Production-Lottery"(sk)]
 
 . stem:[r larr "Epoch-Randomness"(n)]
-. stem:["for " i := 1 " to " sc_n]
-. stem:["    " (pi, d) larr "VRF"(r,i,sk)]
-. stem:["    " A\[i\] larr (d, pi)]
+. stem:["for " s := 1 " to " sc_n]
+. stem:["    " (o, p) larr "VRF"(r,s,e_i,s_k, p_k)]
+. stem:["    " A\[i\] larr (o, p)]
 . stem:["return " A]
 
-For any slot stem:[i] in epoch stem:[n] where stem:[d < r], the block producer
-is required to produce a block. For the definitions of _Epoch-Randomness_
-(<<defn-epoch-randomness>>) and _VRF_ functions (<<sect-vrf>>).
+where stem:["Epoch-Randomness"] is defined in (<<defn-epoch-randomness>>),
+stem:[sc_n] is defined in <<defn-epoch-duration>> , stem:["VRF"] creates the
+BABE VRF transcript (<<defn-babe-vrf-transcript>>) and stem:[e_i] is the epoch
+index, retrieved from the Runtime (<<sect-rte-babeapi-epoch>>). stem:[s_k] and
+stem:[p_k] is the secret key respectively the public key of the authority. For
+any slot stem:[s] in epoch stem:[n] where stem:[d < r], the block producer is
+required to produce a block.
 
 Note that the secondary slots (<<block-production-secondary-slots>>) are running
 along side the primary block production lottery.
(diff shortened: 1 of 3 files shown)


lamafab:
Updated:

image

01_host/05_consensus/block_production.adoc
```diff
@@ -143,6 +150,53 @@ is required to produce a block. For the definitions of _Epoch-Randomness_
...
+where:
+
+* stem:[r] is the Epoch randomness (<<defn-epoch-randomness>>).
+* stem:[s] is the slot number (<<defn-epoch-slot>>).
+* stem:["Enc"_("SC")(...)] encodes its inner value to the corresponding SCALE value.
+* stem:[h(...)] creates a 256-bit Blake2 hash from its inner value.
+* stem:[A_l] is the lengths of the authority list (<<defn-authority-list>>).
+
+To create the necessary VRF outputs, a transcript is construced (<<sect-vrf>>).
+
```
FlorianFranzen, line 203: Till here this all makes sense. I would ask you to rewrite in the style of the primary slot babe algorithm, i.e. like an algorithm that is run by each node at the beginning of an epoch. That then also matches with what you said above, that the assignment is know at the beginning of the epoch, The inputs are the same for all validators, so everybody gets the same result.

This so far I would accept as the spec for a plain, previously known, random round robin style slot assignment.

01_host/05_consensus/block_production.adoc
```diff
@@ -143,6 +150,53 @@ is required to produce a block. For the definitions of _Epoch-Randomness_
...
+p larr sf "dleq_prove"(t_6, i)
+++++
+
+stem:[e_i] is the epoch index (<<defn-epoch-slot>>), stem:[s_k] is the private
+key and stem:[p_k] is the public key of the authority. stem:["dleq_prove"] is
+described in <<defn-vrf-dleq-prove>>. Values stem:[o] and stem:[p] are then used
+in the Pre-Digest item (<<defn-babe-header>>). In case of secondary slots with
+plain outputs, respectively the Pre-Digest being of value _2_, the transcript is
+skipped.
+****
```
FlorianFranzen, line 221: Now this completely breaks with the spec so far. If each validator uses its own private key, the output can not be the same for everybody anymore. Each validator would assume a different ordering, so this does not match with the ordering being known from the start. And VRF are used so the ordering is not revealed. 

I think it is more likely that this is just a second VRF slot, working just like the first one, but on different randomness. 

 At least that is how I understand this section. I have not looked into the code, but I would assume, the same threshold is used her as for the first slot to check if a slot is assigned.

FlorianFranzen (changes requested):
I think the VRF-based slot needs a bit more work. The plain slot looks good.

01_host/05_consensus/block_production.adoc
```diff
@@ -121,26 +128,91 @@ for epoch stem:[cc E_n] and stem:[c in (0, 1)] is the BABE constant
...
+respectively the Pre-Digest being of value _2_, the transcript is skipped.
+****
+
+[#defn-babe-vrf-transcript]
+===== BABE Slot VRF transcript
+****
+The BABE block production lottery requires a specific transcript structure
+(<<defn-vrf-transcript>>). That structure is used by both primary slots
+(<<algo-block-production-lottery>>) and secondary slots
+(<<block-production-secondary-slots>>).
```
FlorianFranzen, line 202: Please mention that they are included in the block via the predigest.

01_host/05_consensus/block_production.adoc
```diff
@@ -280,30 +352,39 @@ stem:[s_(qc)]and expressed in the number of slots.
...
 (<<defn-babe-header>>) as well as the _block signature_
 (<<defn-block-signature>>) as Pre-Runtime and Seal digest items.
 
 [#defn-babe-header]
-.<<defn-babe-header, BABE Header>>
+.<<defn-babe-header, Pre-Digest>>
 ====
-The *BABE Header* of block stem:[B], referred to formally by
-stem:[H_("BABE")(B)] is a tuple and consists of the following components:
+The *Pre-Digest*, or BABE header, of, stem:["P"], is a varying datatype of the
```
FlorianFranzen: Please remove the quotes around `P` here, because they most likely break the PDF build.

01_host/05_consensus/block_production.adoc
```diff
@@ -352,20 +433,11 @@ in which, stem:[E_("id")("BABE")] is the BABE consensus engine unique identifier
...
 ====
 
-[#sect-epoch-randomness]
-==== Epoch Randomness
-At the beginning of each epoch, stem:[cc E_n] the host will receive the
-randomness seed stem:[cc R_(cc E_(n+1))] (<<defn-epoch-randomness>>)
-necessary to participate in the block production lottery in the next epoch
-stem:[cc E_(n+1)] from the Runtime, through the consensus message
-(<<defn-consensus-message-digest>>) in the digest of the first block.
-
```
FlorianFranzen: I felt like this is useful information. Why remove it?

01_host/05_consensus/block_production.adoc
```diff
@@ -352,20 +433,11 @@ in which, stem:[E_("id")("BABE")] is the BABE consensus engine unique identifier
...
 [#defn-epoch-randomness]
 .<<defn-epoch-randomness, Randomness Seed>>
 ====
-For epoch stem:[cc E], there is a 32-byte stem:[cc R_(cc E)] computed
-based on the previous epochs VRF outputs. For stem:[cc E_0] and stem:[cc
-E_1], the randomness seed is provided in the genesis state.
+The epoch randomness is a 32-byte array and is retrieved from the Runtime
+(<<sect-rte-babeapi-epoch>>).
```
FlorianFranzen: Why remove the info about the genesis here?

01_host/06_anv/approval-voting.adoc
```diff
@@ -183,14 +183,13 @@ t_3 larr sf "append"(t_2, "'core'",c_i)\
...
 o = s_k * i\
-(p,x) larr sf "dleq_prove"(t_6, i)
+p larr sf "dleq_prove"(t_6, i)
 ++++
 
-The resulting values stem:[h] and stem:[p] are the VRF input/output pair
-respectively the VRF proof (<<defn-vrf-proof>>). stem:["dleq_prove"] is
-described in <<defn-vrf-dleq-prove>>.
+The resulting value stem:[p] is the VRF proof (<<defn-vrf-proof>>).
+stem:["dleq_prove"] is described in <<defn-vrf-dleq-prove>>.
```
FlorianFranzen, line 192: Normally a VRF proof consists of two component (output + proof), so we should describe both here.

ac_runtime-api/modules/babe.adoc
```diff
@@ -41,10 +41,11 @@ type at genesis will be used. Dynamic slot duration may be supported in the futu
...
 |32-byte array
 
-|_SecondarySlot_
+|_AllowedSlots_
 |Whether this chain should run with secondary slots and wether they are assigned
-in a round-robin manner or via a second VRF.
-| 8bit enum
+in a round-robin manner or via a second VRF (<<sect-block-production-lottery>>).
+|A varying datatype of the following format:
+stem:[{(0,->,"primary slots"),(1,->,"primary & secondary with plain outputs"),(2,->,"primary & secondary with VRF outputs"):}]
```
FlorianFranzen: I am not sure if I agree with the new name and notation. `primary and secondary with plain outputs` sounds like the primary slot changes too. Maybe just leave it as `Secondary Slot`, with zero for none, one for plain and two with VRF.

01_host/05_consensus/block_production.adoc
```diff
@@ -94,6 +95,12 @@ build a block. The stem:[sk] is the block producer lottery secret key and
 stem:[n] is the index of the epoch for whose slots the block producer is running
 the lottery.
 
+In order to ensure consistent block production, BABE uses secondary slots in
+case no authority won the (primary) block production lottery. Unlike the
+lottery, secondary slot assignees are know upfront publically
+(<<block-production-secondary-slots>>). The Runtime dictates what types of
+rounds to execute (<<sect-rte-babeapi-epoch>>).
```
FlorianFranzen: > The Runtime dictates what types of rounds to execute

I am not sure what that could mean. If you mean that the runtime provides information about what slots are currently in use through the babe configuration and consensus messages, it would be best just to link to them here.

01_host/05_consensus/block_production.adoc
```diff
@@ -121,26 +128,91 @@ for epoch stem:[cc E_n] and stem:[c in (0, 1)] is the BABE constant
...
+where stem:["Epoch-Randomness"] is defined in (<<defn-epoch-randomness>>),
+stem:[sc_n] is defined in <<defn-epoch-duration>> , stem:["VRF"] creates the
+BABE VRF transcript (<<defn-babe-vrf-transcript>>) and stem:[e_i] is the epoch
+index, retrieved from the Runtime (<<sect-rte-babeapi-epoch>>). stem:[s_k] and
+stem:[p_k] is the secret key respectively the public key of the authority. For
+any slot stem:[s] in epoch stem:[n] where stem:[d < r], the block producer is
+required to produce a block.
+
+Note that the secondary slots (<<block-production-secondary-slots>>) are running
+along side the primary block production lottery.
```
FlorianFranzen: I would probably say that it runs the secondary slot "as a fallback" for primary block production.

FlorianFranzen (changes requested):
We are getting there, really like the new organization. 

Please also update the longest chain rule with a sentence mentioning it prefers primary slot blocks.

Also, this currently break the PDF build, which will have to be fixed too.

01_host/05_consensus/block_production.adoc
lamafab: I think this is left-over from when the Host had to calculate the randomness seed itself. I now extended it with what functions to call (both at genesis and any further epoch).

lamafab: Respectively, this section is redundant, the retrieval of the epoch randomness is explained in the `defn-epoch-randomness` definition.

01_host/05_consensus/block_production.adoc
lamafab: It doesn't tell us how the value is retrieved from the genesis state. This is covered in the `defn-epoch-randomness` definition.

ac_runtime-api/modules/babe.adoc
lamafab: Updated:

image

01_host/06_anv/approval-voting.adoc
lamafab: `o` is the output. I updated this section which explains how those values are used:

image

ac_runtime-api/modules/babe.adoc
FlorianFranzen: Could you rename the field as well? It seems to be called secondary slot in most code bases as well. then you can remove the `secondary with` as well.

lamafab: Whoops I kind of missed this comment. Do you still want me to change this? I saw that you already made some changes.

01_host/05_consensus/block_production.adoc
FlorianFranzen: The important part here is however that the consensus message received at  `E_n` is to be used for the lottery/assignment in `E_{n+1}`. Might fit better into the Consensus Message section, but it has to be stated somewhere.

ac_runtime-api/modules/babe.adoc
FlorianFranzen: Yes, sorry, I resolved this.

FlorianFranzen:
So, I broke develop yesterday and then I ended up also breaking this branch, when fixing the merge conflict. But it is resolved.

I also ended up dropping a few commits, so we can merge this asap. I will just have to figure out what broke PDF builds. 

You can retrieve all the unmerged commits via a `git pull --rebase`.

01_host/05_consensus/block_production.adoc
FlorianFranzen: Dropped for now. Will be resolved when fixing consensus messages.

01_host/05_consensus/block_production.adoc
FlorianFranzen: Dropped for now, will be resolved with CM fix later.

FlorianFranzen:
I will merge this for now. It will break develop PDF builds on develop, but the build needs to be investigated anyway, due to #495,

PR #496, opened 2022-04-04
[WIP] Syncing Section

lamafab:
Work in progress.

PR #497, opened 2022-04-04
Spec syncing related messages

lamafab:
In favor of #496. Those changes are based on the long(er) discussion in the Spec Underground channel. The information on how to use those messages is quite thin, so let me know what you think.


FlorianFranzen (approved):
Thank you for bringing back that sync section.

PR #494, opened 2022-03-28
Incoperate feedback from parity customer

lamafab:
Fixes #492 .

## Status

 - [x] p.13 2nd paragraph "and is designed and be upgradable" - should be "and is designed to be upgradable"?
   - _This was already changed on `develop`_
 - [x] p.14 definition 1.6 the b's should have subscripted numbers, not exponentials? e.g. see 1.7
 - [x] p. 17 2.1.2 4th paragraph "in a consistent with the Trie structure way"
 - [x] p.25 3.1.2.4 2nd paragraph "any state changes resulting from Runtime enteries"
 - [x] p.25 3.2.1 3rd paragraph "the internals of each intrinsics"
   - _This is already fixed on the latest `develop`
 - [x] p.27 definition 3.6 "The header of block B, Head(B)" - there is already a definiton for "Head(B)" in definition 1.13 - this confused me.
   - _We kind of define `B := Head(C)`, head of chain `C`, but never actually use it? At the same time we define `Head(B)` to be the block header of block `B`. So this is indeed confusing. I guess we can remove `B := Head(C)` or change it?_
 - [x] p.29 alg 3.4 I believe "Set-Storage-State-At" is actually supposed to be "Set-State-At" ? For instance, see Definition 3.10 same page.
 - [x] p.36 4th bullet "to perform information about blocks" - should be request, maybe?
 - [x] p.48 def 6.11 confused me, "c ∈ (0,1)" - I did not grok that as the range of 0 to 1 but as the set of 0,1 "{0,1}?". This is probably a notational failure on my end but perhaps I'm not unique.
   - This is kind of confusing, indeed. Rust code: `pub c: (u64, u64)`. I adjusted this to: "[...] It’s expressed as a rational, (𝑥,𝑦), where 𝑥 is the numerator and 𝑦 is the denominator.".
 - [x] p.49 3rd paragraph from note 6.15 - "formal security analysis" is a pretty big claim, should back that up with a link to said analysis.
   - _Referenced the W3F research paper on BABE_.
- [x] p.64 7.2 Preliminaries - "The SCALE codec is defined in a separate document known as ``The Polkadot Host - Protocol Specification''. - Infact, that's the very document we're reading, should probably link to the appendix B.
  - _This was already changed on `develop`_

PR #499, opened 2022-04-07
Small fix for 494

lamafab:
Small changes for PR #494


FlorianFranzen (approved):
Weird, I feel like I merged some of these before. But dealing with all the Grandpa stuff was quite messy.

PR #500, opened 2022-04-07
Fix block header digest item and consensus message

lamafab:
As discussed. Note that I'll have a meeting with Andre (Parity) and ChainSafe soon, so there will (probably) be upcoming changes regarding the scheduled/forced authority changes. I also added BEEFY to the consensus message definition, which is quite thin tbh.

**Header Digest**

image

**Consensus message** (we might want to split this into multiple definitions):

image


FlorianFranzen (approved):
Thank you, looks great.

PR #502, opened 2022-05-05
Incoperate recommendations from Gossamer Audit

### LINKED ISSUE #493: Incoperate recommendations from Gossamer Audit
There have been a lot of good recommendations in the gossamer audit. I will make issues out of them, so that we can update our spec accordingly.

lamafab:
Fixes #493. This point is still work-in-progress, or rather, undecided:

>  Definition 6.12 is insufficient in clarifying how these non-integer computations should be performed - Doesn't specify floating points or how it should be implemented - so rounding and inconsistent results might cause a consensus split (i.e. when the rounding occurs) - is this floating point? fixed decimal? etc


FlorianFranzen (approved):
Looks good. I would be a bit more stern on Scale's upper bound. I might even put it in a `WARNING` block. We should clearly states that the size of the encoded scale should serve as a strict upper bound for any size-encoded structures decoded from it. If a structure exceeds this size, this should be considered an encoding error, to prevent denial of service attacks.

PR #505, opened 2022-05-06
Host: State: Node header fixes

qdm12:
- Fix possible typo `length of the partial key length` to `length of the partial key`
- Remove `in 4-bit nibbles` since the node variant can be expressed using 2 to 8 bits, and the partial key length in 2 to 8 bits + N*8 bits

Let me know if I misunderstood anything, I would happy to re-clarify my understanding or the text :+1:


FlorianFranzen (approved):
Thanks!

PR #506, opened 2022-05-06
Update forced changes

lamafab:
A small update regarding forced changes. I think this is sufficient.


lamafab:
FYI: I'm currently working on a chart, will be added to this PR

lamafab:
Updated

lamafab:
## Printscreens

image

image

FlorianFranzen (approved):
Content-wise looks great. I think this is an important recovery feature, which can only work if properly supported by all clients, so thank you for charts!

Could you try exporting the charts from draw.io as svg instead? Then we do not need the separate drawio file and can edit them directly.

lamafab:
Done. It now uses SVG, I also made some adjustments to the chart itself, such as increasing sizes. Btw that `.drawio` file was meant to be temporary in the directory root, forgot to remove it. I now moved it into the `asset/` directory.

PR #511, opened 2022-05-26
Host: State: Node headers containing hashes

qdm12:
Correct if I'm wrong but from what I understand, these two new node headers for leaves and branches mean their value is hashed, not that they contain hashes (especially plural hashes)?

### DIFF
--- a/01_host/02_state/state_storage_trie.adoc
+++ b/01_host/02_state/state_storage_trie.adoc
@@ -269,8 +269,8 @@ v = {
     (01, "Leaf", p_l = 2^6),
     (10, "Branch Node with" k_N !in cc K, p_l = 2^6),
     (11, "Branch Node with" k_N in cc K, p_l = 2^6),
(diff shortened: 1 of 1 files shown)


lamafab:
Correct, it refers to this: https://spec.polkadot.network/#defn-node-hashes

image

I.e. it contains _the_ hash of the subvalue if it's larger than 32 bytes (rather than the subvalue itself).

lamafab:
@qdm12 I recommended some changes: https://github.com/qdm12/polkadot-spec/pull/1

FlorianFranzen:
@qdm12 Thank you for your contribution. I will merge this and then @lamafab additions.

qdm12:
@lamafab sorry for the delay, I think it got merged with the initial PR content, and not your recommended changes.
I'll make another PR to correct the correction 😉 

EDIT: never mind, I saw you merged as #514 awesome 👍

PR #518, opened 2022-07-06
Fix state trie `version` to be as value instead of `pointer-size`

qdm12:


tomaka:
If I'm not mistaken, the `version` is indeed not a `pointer-size`, but it's not a `pointer` either. It's simply a `i32` that specifies the version, either 0 or 1.

qdm12:
I updated the changes so it's just the `state version` and no longer a pointer nor pointer-size.

Issue #527, opened 2022-08-19
`/ipfs/id/1.0.0` is not *really* a request-response protocol


tomaka:
The spec mentions that in a request-response protocol, the initiator sends a single message containing the request and that this request is prefixed with its length.
This is indeed accurate, for all protocols except for `/ipfs/id/1.0.0` (because it was designed by Protocol Labs and not Parity).

For `/ipfs/id/1.0.0`, the initiator doesn't send any message at all, not even a `0`. The sending side of the initiator of the substream can be immediately closed.

lamafab:
Fixed in #529

PR #529, opened 2022-08-24
Incorporate Parity feedback

### LINKED ISSUE #524: Incorporate feedback from Parity
As discussed in our matrix channel. Following points are to consider:

* Fully deprecate Changes Tries.
* Consider renaming "Runtime Entries" to "Entrypoint".
* Look into SCALE encoding exceptions when returning primitives types from Runtime.
* Add deprecation notice about `Core_version`, spec how that version can be read directly from Wasm custom section.
* Spec malloc and free?

lamafab:
Fixes #524 and #527.

Regarding:

> Look into SCALE encoding exceptions when returning primitives types from Runtime.

As can be seen in our Underground channel and my source code analysis there, it appears like the current spec which says that every return value from the Runtime is a `u64` indicating pointer+size is indeed correct.

Additionally, `malloc` and `free` is not specced in this PR. Not sure whether it should.


01_host/03_transition/02_extrinsics.adoc
```diff
@@ -161,8 +161,8 @@ be included in the current block as explained in <<algo-build-block>>.
 |Unix epoch time (<<defn-unix-time>>)
 
 |uncles00
-|Array of block headers
-|Provides a list of potential uncle block headers (<<defn-block-header>>) for a given block
+|Array of block headers (auxiliary information)
+|Provides a list of potential uncle block headers (<<defn-block-header>>) for a given block, if any.
```
bkchr: Uncles are not used anymore and I think they never were. Meaning you don't need to pass the uncles.

01_host/03_transition/01_runtime-interaction.adoc
```diff
@@ -130,54 +133,43 @@ of 64KB. This memory should be made available to the Polkadot Runtime
...
-<<sect-managing-multiple-states>>.
+[#sect-runtime-version-custom-section]
+===== Runtime Version Custom Section
+
+For newer Runtimes, the Runtime version (<<defn-rt-core-version>>) can be read
+directly from the Wasm custom section with the name `runtime_version`. The
+content is a SCALE encoded structure as described in <<defn-rt-core-version>>.
+
+Retrieving the Runtime version this way is preferred over calling the
+`Core_version` entrypoint since it involves significantly less overhead.
```
FlorianFranzen, line 177: I think it would make mores sense to add this to the section for runtime constants in the Runtime API chapter.

FlorianFranzen (changes requested):
Thanks for integrating all of these. Looks good apart from where you added the info about the `runtime_version` constant.

01_host/03_transition/01_runtime-interaction.adoc
lamafab: I'm not sure I agree, it's not exactly a constant exposed by the Runtime/Wasm but the entry of the Wasm custom section. I think having it in the "Interacting with the Runtime" section is appropriate.  (I updated this PR with a link to the Wasm custom section spec, just FYI)

FlorianFranzen: Ok, then I missunderstood. Perfect, thanks.

PR #531, opened 2022-09-07
extrinsics: Clarify bit positioning

lexnv:
We have noted that the first bit of the extrinsic might cause confusion.

To clarify the documentation, use MSB instead to indicate the signed
or unsigned extrinsic payload.


niklasad1:
@lamafab @FlorianFranzen would be great if you could review this one :)

PR #533, opened 2022-09-08
Update spec for PeerId

### LINKED ISSUE #530: Precisely define the format of PeerId
I think that would not be a bad idea to precisely define how to convert between PeerId and ed25519 public key in the spec, rather than linking to the corresponding libp2p specification.

Libp2p allows PeerIds to use the CIDv1 format, which Substrate doesn't support, and allows public keys to be something else than ed25519, which, again, Substrate doesn't support. In other words, we're kind of not-really-conforming to the libp2p spec. But at the same time it's hard to say whether we're conforming, because the libp2p spec is really ambiguous about what is required and what isn't, and the libp2p people don't seem to have the same definition of the word "specification" as the rest of the IT industry.

Furthermore, the libp2p specification requires an implementation of protobuf while at the same time requiring the fields to be in a specific order, which kind of defeats the point of defining the format as a protobuf struct instead of just bytes. I think that using bytes is just easier for everyone.

Anyway, the format of a PeerId is as follows: `[0, 36, 8, 1, 18, 32, <32 bytes of ed25519 public key>]`

A small description of each byte:

- `0`: indicates that the content of the multihash is not hashed.
- `36`: length of the content of the multihash.
- `8`: field 1 of the public key protobuf of type number
- `1`: value of field 1, indicates that this is ed25519. If we ever add other kind of cryptographic keys, this is the byte that would need to change.
- `18`: field 2 of the public key protobuf of type bytes
- `32`: length of field 2 of the protobuf
- (the rest): actual content of field 2 of the protobuf

lamafab:
Fixes #530 

image

### DIFF
--- a/01_host/04_networking/01_fundamentals.adoc
+++ b/01_host/04_networking/01_fundamentals.adoc
@@ -93,7 +93,7 @@ the used key type] (field stem:[1] of value stem:[1] implies _ED25519_).
 * stem:[b_4], stem:[b_5] and stem:[b_(6..37)] are a protobuf encoded field-value
 pair where stem:[b_5] indicates the length of the public key followed by the the
 raw ED25519 public key itself, which varies for each Polkadot Host and is always
-32 bytes (field stem:[2] is of value stem:[32] + public key).
+32 bytes (field stem:[2] contains the public key, which has a field value length prefix).
 ====
 
 [#sect-discovery-mechanism]
--- a/01_host/04_networking/01_fundamentals.adoc
+++ b/01_host/04_networking/01_fundamentals.adoc
@@ -61,9 +61,39 @@ course of the discovery mechanism described in <<sect-discovery-mechanism>>.
 [#defn-peer-id]
 .<<defn-peer-id,PeerId>>
 ====
-The Polkadot node’s _PeerId_, formally referred to as stem:[P_(id)], is derived
-from the ED25519 public key and is structured as defined in the
-https://docs.libp2p.io/concepts/peer-id/[libp2p specification].
+The Polkadot node’s PeerId, formally referred to as stem:[P_(id)], is derived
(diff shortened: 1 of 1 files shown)


tomaka (approved):
Note that I don't necessarily think that we should explain the meaning of each byte. I explained that in the issue for the sake of being informative, but I don't really think that this should be part of a spec.

lamafab:
I think we should leave it there. When we make claims that the PeerId is based on the libp2p spec but does not fully conform, then it's probably useful to specify how it's different respectively what the values mean exactly.

01_host/04_networking/01_fundamentals.adoc
```diff
@@ -61,9 +61,39 @@ course of the discovery mechanism described in <<sect-discovery-mechanism>>.
...
+* stem:[b_0] is the https://github.com/multiformats/multihash#multihash[multihash prefix]
+of value stem:[0] (implying no hashing is used).
+* stem:[b_1] the length of the PeerId (remaining bytes).
+* stem:[b_2] and stem:[b_3] are a single protobuf encoded field-value pair
+https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#keys[indicating
+the used key type] (field stem:[1] of value stem:[1] implies _ED25519_).
+* stem:[b_4] and stem:[b_5] are a single protobuf encoded field-value pair
+indicating the length of the public key (field stem:[2] is of value stem:[32]).
+* stem:[b_(6..37)] is the raw ED25519 public key, which varies for each Polkadot
+Host and is always 32 bytes.
```
tomaka: I haven't caught this, but for clarification: bytes 6 to 37 are the value of field 2 of the protobuf. Byte 5 (value 32) indicates the length of the value of the field.

FlorianFranzen (approved):
Looks good, thanks!

PR #535, opened 2022-09-14
changed return type of ext_storage_root to i64 (previously i32) 

bhargavbh:
Double checked for other storage, child_storage, and trie functions in Host API. Deciding how we handle the fact that Rust types are converted into FFI types u32/u64 (which are not primitive type in WASM) is open for discussion.

### DIFF
--- a/ab_host-api/storage.adoc
+++ b/ab_host-api/storage.adoc
@@ -187,12 +187,12 @@ the 256-bit Blake2 storage root.
 ===== Version 2 - Prototype
 ----
 (func $ext_storage_root_version_2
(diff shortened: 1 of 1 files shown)


bkchr:
> Deciding how we handle the fact that Rust types are converted into FFI types u32/u64 (which are not primitive type in WASM) is open for discussion

TBH, I still don't understand what is so complicated? I mean we can just document how a u64 looks bitwise and then it should be really easy to find out how to get a u64 out of a i64.

FlorianFranzen (approved):
Congrats to passing your interview question twice. :wink:

PR #537, opened 2022-09-14
Reference EdDSA RFC for ed25519 signatures and explicitly require

drskalman:
Zebra verification


burdges:
I bet the zip covers everything via it's references but..

You might reference HdV's blog post for the explanation of zebra.  https://hdevalence.ca/blog/2020-10-04-its-25519am

burdges (approved):

FlorianFranzen (approved):
Looks good, thank you.

PR #558, opened 2022-10-11
Change to CC BY-SA license

lamafab:
The current GPL license is basically left-over from when we had tests in this repo


lamafab:
As discussed in the meeting, we're going with the GPL/CC dual-license (for code and spec, respectively)

PR #566, opened 2022-10-19
Mention legality of calling functions multiple times in a row

### LINKED ISSUE #564: Make it explicit that it is legal to call ext_storage_start_transaction multiple times in a row
It is legal to start nested transactions, in other words for example call `start_transaction` twice in a row then `commit_transaction` twice in a row.
While the spec doesn't say that it's illegal to do so and uses the words `nested` and `the last transaction`, it's a bit ambiguous, and it is IMO far from being obvious that is indeed legal. Thus I would suggest to explicitly mention that it's legal to do so.

lamafab:
Fixes #564


FlorianFranzen (approved):
Looks good. Thanks!

PR #576, opened 2022-11-02
Improve the language when duplicate libp2p key

tomaka:
PR does two things:

- Replace "undefined behavior" with "unspecified behavior". I understand that you mean "the behavior can be anything", but the words "undefined behavior" are really scary in IT. The behavior is not actually "undefined", as the node still does a specific thing, it's just that what it does might not conform to the spec.
- Remove " and can result in equivocation", because that's not actually true. Right now you can never be equivocated for using the same libp2p key multiple times. Even if this could result in an equivocation somehow, this is covered by the words "unspecified behavior".

### DIFF
--- a/01_host/04_networking/01_fundamentals.adoc
+++ b/01_host/04_networking/01_fundamentals.adoc
@@ -51,12 +51,12 @@ Each Polkadot Host node maintains an ED25519 key pair which is used to
 identify the node. The public key is shared with the rest of the network
 allowing the nodes to establish secure communication channels.
 
-Each node must have its own unique ED25519 key pair. When two or more nodes use
+Each node must have its own unique ED25519 key pair. If two or more nodes use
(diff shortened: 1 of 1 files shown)


lamafab:
> Remove " and can result in equivocation", because that's not actually true. Right now you can never be equivocated for using the same libp2p key multiple times. Even if this could result in an equivocation somehow, this is covered by the words "unspecified behavior".

Well, when using two keys at the same time it's quite probable that you'll send two contradictory GRANDPA votes, for example.

tomaka:
> Well, when using two keys at the same time it's quite probable that you'll send two contradictory GRANDPA votes, for example.

A GRANDPA equivocation happens if you send two votes signed with the same GRANDPA keys. We don't care at all about the identity of the node which has sent the votes. In other words, this is completely orthogonal to the networking key.

Issue #577, opened 2022-11-03
ext_storage_get_version_1 behavior w.r.t. child tries


tomaka:
When reading the location of a child trie using `ext_storage_get_version_1`, the trie root hash of the child trie should be returned.

But there's a twist: the returned hash only takes into account the child trie writes that have been performed until the last time `ext_storage_child_storage_root_version_1` was called.

On the other hand, `ext_storage_child_storage_root_version_1` always recalculates the trie root hash of the child trie and takes into account all the writes that have been performed.

So for example if you call `ext_storage_child_set_version_1`, then `ext_storage_child_storage_root_version_1`, then `ext_storage_child_set_version_1` again, then read the child trie root hash using `ext_storage_get_version_1`, it returns the hash that `ext_storage_child_storage_root_version_1` returned.

Issue #540, opened 2022-10-04
Corner cases about child tries are unexplained


tomaka:
Some questions are in my opinion not really clear in the spec:

- What happens if the runtime calls the `ext_storage_get_version_1` or `ext_storage_read_version_1` host function with a key K, where K is in fact a child trie? As far as I understand, this would return the trie root hash of the child trie, but this isn't really clear in the spec.
- What happens if the runtime calls the `ext_storage_set_version_1` host function with a key K, where K is in fact a child trie? I assume that it destroys the child trie, but this isn't really clear.
- What happens if the runtime calls any of the `ext_default_child_storage_*` host functions with a `child_storage_key` that is in fact not a child trie? I assume that this is the equivalent of a runtime panic?

lamafab:
Fixed in #554

lamafab:
Test coverage in https://github.com/w3f/polkadot-tests/pull/241, should be merged, soon

tomaka:
I don't see how https://github.com/w3f/polkadot-spec/pull/554 covers any of the questions that I've raised, and I don't see any answer to them in the spec.

For some reason I don't have a "reopen" button, but if I had one I'd reopen the issue.

tomaka:
I would say that the first two points are now covered by https://github.com/w3f/polkadot-spec/issues/577 and https://github.com/w3f/polkadot-spec/issues/575

Only remains:

> What happens if the runtime calls any of the ext_default_child_storage_* host functions with a child_storage_key that is in fact not a child trie? I assume that this is the equivalent of a runtime panic?

PR #580, opened 2022-11-07
[WIP] Describe implementation assumptions of Substrate regarding (child) storages

### LINKED ISSUE #575: Document writing/clear_prefix behavior w.r.t. child tries
It should be documented that:

- Any call to `ext_storage_set_version_1` with a key that starts with `:child_storage:` is silently ignored.
- Any call to `ext_storage_clear_prefix` (I guess both versions?) with a key that "clashes" with `:child_storage:` is silently ignored.

"Clashes" means that for example `clear_prefix(":")`, `clear_prefix(":child_sto")`, or `clear_prefix(":child_storage:foobar")` are all silently ignored, but `clear_prefix(":foo")` is not ignored.

References:

- For storage_set: https://github.com/paritytech/substrate/blob/c8653447fc8ef8d95a92fe164c96dffb37919e85/primitives/state-machine/src/ext.rs#L392
- For clear_prefix: https://github.com/tomaka/polkadot/blob/5ae005c244295d23586c93da43148b2bc826b137/primitives/state-machine/src/ext.rs#L470-L476 and https://github.com/tomaka/polkadot/blob/5ae005c244295d23586c93da43148b2bc826b137/primitives/storage/src/lib.rs#L229-L235

lamafab:
Fixes #575  and #577 (#540 should be later completed here, too).

@tomaka I would appreciate if you could quickly double check this, especially my claim on how child storage keys are constructed and that `get` returns _None_ if `child_storage_root` has never been called. Those are random guesses of mine :)

image

### DIFF
--- a/ab_host-api/storage.adoc
+++ b/ab_host-api/storage.adoc
@@ -3,6 +3,7 @@
 
 Interface for accessing the storage from within the runtime.
 
+[#sect-storage-assumptions]
 ==== Implementation Assumptions
 The storage and child storage (<<sect-child-storage-api>>) API in the Substrate
 implementation makes some behavioral assumptions on the underlying storage
@@ -15,26 +16,33 @@ child storage. However, the storage API described in this section and the child
 storage API share the same database. This means that the storage API can
 retrieve child storage entries.
 
-For example, calling `ext_storage_get_version_1` with the key
-`:child_storage:default:some_child:some_key` is equivalent to calling
-`ext_default_child_storage_get_version_1` with child storage key `some_child`
-and key `some_key`.
+For example, calling `ext_storage_get_version_1` (<<sect-ext-storage-get>>) with
+the key `:child_storage:default:some_child:some_key` is equivalent to calling
+`ext_default_child_storage_get_version_1`
+(<<sect-ext-default-child-storage-get>>) with child storage key `some_child` and
+key `some_key`.
 
 Importantly, the storage API can only _read_ child storage entries, but must
 implement write protection. Any function that modifies data must silently ignore
-any keys that start with the `:child_storage:` prefix or any keys that are a
-substring of that prefix, such as `:child_sto` (but not `:other`, for example).
-If the function expects a return value, then _None_ (<<defn-option-type>>)
-should be returned.
+any keys that start with the `:child_storage:` prefix. For
+`ext_storage_clear_prefix` (<<sect-ext-storage-clear-prefix>>) specifically,
+this also applies to any key that is a substring of `:child_storage:` prefix,
+such as `:child_sto` (but not `:other`, for example). If the function expects a
+return value, then _None_ (<<defn-option-type>>) should be returned.
 
 Additionally, calling `ext_storage_get_version_1` on a child storage key
 directly (without a key within the child storage), such as
-`:child_storage:defaut:some_child`, retuns the root of the child storage. This
-is described in TODO.
+`:child_storage:defaut:some_child`, the function returns the root of the child
+storage from when `ext_default_child_storage_root`
+(<<ext-default-child-storage-root>>) _has been last called_ or _None_
+(<<defn-option-type>>) if it has never been called yet. Respectively, that root
+value is cached. Calling `ext_default_child_storage_root` directly always
+recomputes the current root value.
 
-See the https://github.com/w3f/polkadot-spec/issues/575[polkadot-spec issue
-#575] and the https://github.com/paritytech/substrate/issues/12461[substrate
-issue #12461] for more info.
+See the following issues for more information:
+* https://github.com/w3f/polkadot-spec/issues/575
+* https://github.com/w3f/polkadot-spec/issues/577
+* https://github.com/paritytech/substrate/issues/12461
 
 [#defn-state-version]
 .<<defn-state-version, State Version>>
@@ -56,8 +64,8 @@ merkle proof (<<defn-hashed-subvalue>>).
 ==== `ext_storage_set`
 Sets the value under a given key into storage.
 
-IMPORTANT: Please note the section disclaimer (<<sect-storage-api>>) regarding
-child storage keys.
+IMPORTANT: Do note that this API must implement some specific behaviors,
+described further in <<sect-storage-assumptions>>
 
 ===== Version 1 - Prototype
 ----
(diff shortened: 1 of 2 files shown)


ab_host-api/storage.adoc
```diff
@@ -3,11 +3,46 @@
...
-If the function expects a return value, then _None_ (<<defn-option-type>>)
-should be returned. See
-https://github.com/paritytech/substrate/issues/12461[substrate issue #12461].
+[#sect-storage-assumptions]
+==== Implementation Assumptions
+The storage and child storage (<<sect-child-storage-api>>) API in the Substrate
+implementation makes some behavioral assumptions on the underlying storage
+architecture. While Polkadot Host implementers can decide for themselves on how
+those APIs are implemented, those behaviors must be replicated in order for the
+Runtime to be executed deterministically. In Substrate, child storages are
```
tomaka, line 12: I don't understand the purpose of this paragraph. You seem to be literally describing what a specification is.

ab_host-api/storage.adoc
```diff
@@ -3,11 +3,46 @@
...
+namespaced respectively segregated by attaching the `:child_storage:default:`
+prefix to every child storage keys, followed by `:<KEY>` for the key within that
+child storage. However, the storage API described in this section and the child
+storage API share the same database. This means that the storage API can
+retrieve child storage entries.
+
+For example, calling `ext_storage_get_version_1` (<<sect-ext-storage-get>>) with
```
tomaka, line 19: Why is this not explained under `ext_storage_get_version_1`?
And why "for example"? We want a precise list, not examples?

ab_host-api/storage.adoc
```diff
@@ -3,11 +3,46 @@
...
+the key `:child_storage:default:some_child:some_key` is equivalent to calling
+`ext_default_child_storage_get_version_1`
+(<<sect-ext-default-child-storage-get>>) with child storage key `some_child` and
+key `some_key`.
+
+Importantly, the storage API can only _read_ child storage entries, but must
+implement write protection. Any function that modifies data must **silently ignore**
+any keys that start with the `:child_storage:` prefix. For
+`ext_storage_clear_prefix` (<<sect-ext-storage-clear-prefix>>) specifically,
```
tomaka, line 28: Similarly, why not explain this under `ext_storage_clear_prefix`?

ab_host-api/storage.adoc
```diff
@@ -3,11 +3,46 @@
...
+prefix to every child storage keys, followed by `:<KEY>` for the key within that
+child storage. However, the storage API described in this section and the child
+storage API share the same database. This means that the storage API can
+retrieve child storage entries.
+
+For example, calling `ext_storage_get_version_1` (<<sect-ext-storage-get>>) with
+the key `:child_storage:default:some_child:some_key` is equivalent to calling
+`ext_default_child_storage_get_version_1`
+(<<sect-ext-default-child-storage-get>>) with child storage key `some_child` and
+key `some_key`.
```
tomaka, line 23: Except that you say it's not equivalent two paragraphs later

ab_host-api/storage.adoc
```diff
@@ -3,11 +3,46 @@
...
+storage from when `ext_default_child_storage_root`
+(<<ext-default-child-storage-root>>) has been **last called** or _None_
+(<<defn-option-type>>) if it has never been called. Respectively, that root
+value is cached. Calling `ext_default_child_storage_root` directly always
+recomputes the current root value.
+
+See the following issues for more information:
+* https://github.com/w3f/polkadot-spec/issues/575
+* https://github.com/w3f/polkadot-spec/issues/577
+* https://github.com/paritytech/substrate/issues/12461
```
tomaka, line 46: We shouldn't be linking issues in a specification?

ab_host-api/storage.adoc
```diff
@@ -3,11 +3,46 @@
...
+those APIs are implemented, those behaviors must be replicated in order for the
+Runtime to be executed deterministically. In Substrate, child storages are
+namespaced respectively segregated by attaching the `:child_storage:default:`
+prefix to every child storage keys, followed by `:<KEY>` for the key within that
+child storage. However, the storage API described in this section and the child
+storage API share the same database. This means that the storage API can
+retrieve child storage entries.
+
+For example, calling `ext_storage_get_version_1` (<<sect-ext-storage-get>>) with
+the key `:child_storage:default:some_child:some_key` is equivalent to calling
```
tomaka, line 20: I don't think that this is accurate. As far as I understand, you can only call `ext_storage_get_version_1` on `:child_storage:default:some_child` in order to get the root hash of the child trie, but you can't get the keys of the child trie.

ab_host-api/storage.adoc
lamafab: So we've been telling the implementers that the "main" storage is fully separated from the child storages and that this functionality can be implemented however they want. But with this new specification we're basically saying that those storages are really not that separate, meaning we kind of insist now that child storage entries are namespaced by a specific prefix and that the Storage API actually _can_ access Child Storages. This paragraph is meant to imply that the Host implementation must simply deal with this "ugliness", but I guess I can clarify this further.

Regarding your other points.

> Why is this not explained under ext_storage_get_version_1?
> And why "for example"? We want a precise list, not examples?

I put everything into one section so it's clear how the underlying data is structured in the Substrate implementation. Those things must also be considered for `read`, `append`, `next_key`, etc, too. But I should replace the "for example" there with something more formal.

> Similarly, why not explain this under ext_storage_clear_prefix?

As mentioned above, it's to keep things packed together; functions that modify data simply check whether the key starts with `:child_storage:`, but the `clear_prefix` function makes an exception by also checking whether the input is a substring of `:child_storage:`. If you think it's more appropriate to move that section to `ext_storage_clear_prefix` I can do so.

tomaka: > So we've been telling the implementers that the "main" storage is fully separated from the child storages and that this functionality can be implemented however they want. But with this new specification we're basically saying that those storages are really not that separate, meaning we kind of insist now that child storage entries are namespaced by a specific prefix and that the Storage API actually can access Child Storages. This paragraph is meant to imply that the Host implementation must simply deal with this "ugliness", but I guess I can clarify this further.

I still don't understand. You *told* something to the implementers, and because you told them something you're writing the spec in a different way than you would have if you hadn't told them?

ab_host-api/storage.adoc
tomaka: This spec isn't supposed to be an explanation of how Substrate works. It is Substrate that is supposed to conform to what this spec says. Linking to an issue saying that we want to remove the feature is IMO not only inappropriate (it places a judgement over how it works) but also doesn't remove the fact that child tries will continue to have to be supported and will forever be part of the Polkadot spec even if we remove its usage from Substrate.

Issue #590, opened 2022-11-24
RFC: Add a way to ask for a storage proof of just the hash of a value


tomaka:
# Context and motivation

In the Polkadot network, there are two kinds of nodes: full nodes, that have the storage of recent blocks, and light clients, that don't. In order for light clients to work, they need to request from full nodes the storage items that they want to access. To do so, they send a networking request.

When it receives such networking request, the full node returns a Merkle proof containing the values that have been requested.

In some situations, the light client doesn't care about the value of the storage items that it wants to access, but only their hash. What this RFC proposes is to add a way for the light client to indicate, in the networking request, that it doesn't care about the values of the storage items that it requests but only their hash. When this is the case, the full node can omit the storage values from the Merkle proof that it sends back, and thus reduce its size, sometimes considerably.

This would be helpful, for example:

- In order to check whether a storage item still matches a locally-known value without downloading it. This is important for big storage items such as `:code`.
  - In particular, when the smoldot light client connects to a chain, almost all of its initialization time and bandwidth are used to download `:code` from a full node. By implementing this proposal, the light client could store the value of `:code` on the disk and check if it is still the same as on chain instead of always downloading it.
- In order to enumerate the list of keys in the storage, in order to answer the JSON-RPC requests that require you to enumerate keys.

I won't go in details as to how values can be omitted from Merkle proofs, as this is a bit complicated and out of scope here, however it is important to know that in order for values to be omitted, a chain must have transitioned to the "version 1" of the trie format by switching the `state_version` field in the runtime spec to `1`. If it hasn't done this, it is not possible to omit storage values, as otherwise the proof would no longer be verifiable.
Chains that are still using version 0 of the trie format should still implement this proposal, but they would leave the values in the proof like is currently the case.
It is also technically possible for a chain to be in the middle of a version 0 -> version 1 trie format transition, but this isn't problematic either, as values can be left in the Merkle proof when the corresponding storage item is still at version 0.

# Changes

The current Protobuf schema for the "light client requests" protocol can be found here: https://github.com/paritytech/substrate/blob/08d1b2c7846f0280aa57dc1145a2c631b12c0789/client/network/light/src/schema/light.v1.proto

This RFC suggests the following diff:

```diff
@@ -11,6 +11,7 @@ message Request {
                RemoteReadRequest remote_read_request = 2;
                RemoteReadChildRequest remote_read_child_request = 4;
                // Note: ids 3 and 5 were used in the past. It would be preferable to not re-use them.
[...]

PR #591, opened 2022-11-24
[Early Review] Version 2 of light client request messages

lamafab:
Regarding #590. I generally still need to to describe that "read proof" in detail, it's not quite clear what it consists off or how it's verified/handled.

### DIFF
--- a/01_host/04_networking/10_light-client.adoc
+++ b/01_host/04_networking/10_light-client.adoc
@@ -48,7 +48,7 @@ Where the `request` can be one of the following fields:
 
 |`RemoteReadChildRequestV2`
 |7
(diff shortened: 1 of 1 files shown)


bhargavbh:
LGTM. The verification of a response against the merkle proof is planned to be covered in the Light Client Chapter

tomaka:
https://github.com/w3f/polkadot-spec/issues/590 was a proposal for changes that I've opened here because we don't have any other place where to open this kind of issues. It's a "Request For Comment". It was very much not supposed to make it into the spec like that.

lamafab:
_Reverted_

Issue #588, opened 2022-11-23
RFC: Change `Storage::clear_prefix`, `DefaultChildStorage::clear_prefix` and  `DefaultChildStorage::storage_kill`


bkchr:
# Proposal

We would like to add the third version of the `clear_prefix` function for the top trie and child trie:
```
fn Storage_clear_prefix_version_3(u64, u64, u64) -> u64;
fn DefaultChildStorage_clear_prefix_version_3(u64, u64, u64) -> u64;
```

We also would like to add the third version of `storage_kill` for the child trie:
```
fn DefaultChildStorage_storage_kill_version_3(u64, u64, u64) -> u64;
```

To make the function better understandable, here the high level rust representation of the function:

```
fn clear_prefix(prefix: &[u8], maybe_limit: Option<u32>, maybe_cursor: Option<Vec<u8>>) -> MultiRemovalResults;

pub struct MultiRemovalResults {
	/// A continuation cursor which, if `Some` must be provided to the subsequent removal call.
	/// If `None` then all removals are complete and no further calls are needed.
        /// Must be the key in the state after the last deleted key.
	pub maybe_cursor: Option<Vec<u8>>,
	/// The number of items removed from the state.
        /// 
        /// Keys are are already in the overlay do not count towards keys being removed from state. E.g. the overlay already has
       /// key `AB`, the state also has `AB` and you are deleting with `prefix` `A`. `AB` would not be counted for `state`.
	pub state: u32,
	/// The number of unique keys removed, taking into account both the state and the overlay.
	pub unique: u32,
	/// The number of processed keys from the state. Should in maximum be `limit`.
	pub loops: u32,
}
```

The important changes being here the introduction `maybe_cursor` and the change of the return value to `MultiRemovalResults`.

`DefaultChildStorage::storage_kill` only differs in the way that it doesn't provide any `prefix` to `clear_prefix`, besides that the implementation is exactly the same.

## Implementation of `clear_prefix`

Clear keys from the state and the overlay of the currently modified keys that are starting with `prefix`. If there is no `maybe_limit` set, all keys under the given `prefix` are deleted. If `maybe_limit` is set, in maximum `limit` keys are being deleted from the state. The keys that are deleted in the overlay aren't counted towards the limit. If `maybe_cursor` is given, the implementation needs to seek the given `cursor` and start the iteration with `cursor` (or the first key in lexicographical order after `cursor` if not present). Returns `MultiRemovalResults` to inform the caller about the operation.

[...]

PR #592, opened 2022-11-25
[PoC/WIP] Spec version 3 of clear_prefix and kill_storage

lamafab:
Context: #588 (Not merging until the RFC has been finalized)

- [x] ext_storage_clear_prefix_version_3
- [x] ext_default_child_storage_clear_prefix_version_3
- [x] ext_default_child_storage_kill_storage_version_4

### DIFF
--- a/ab_host-api/child_storage.adoc
+++ b/ab_host-api/child_storage.adoc
@@ -209,8 +209,54 @@ k = {(0,-> c),(1,-> c):}
 ++++
 +
 where _0_ indicates that all keys of the child storage have been removed,
-followed by the number of removed keys, stem:[c]. The variant _1_ indicates that
-there are remaining keys, followed by the number of removed keys.
+followed by the unsigned 32-bit number of removed keys, stem:[c]. The variant
+_1_ indicates that there are remaining keys, followed by the unsigned 32-bit
+number of removed keys.
+
+===== Version 3 - Prototype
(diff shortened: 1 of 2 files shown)


lamafab:
This is now implemented in the Substrate codebase.

tomaka:
> This is now implemented in the Substrate codebase.

Do you have a link to a PR? I can't find it.

If it's indeed been implemented, it arguably shouldn't have been without a RFC/discussion.

lamafab:
I found this by using `git blame`: https://github.com/paritytech/substrate/pull/11490

lamafab:
So I found this: https://github.com/paritytech/substrate/pull/11490 Looks like it has been there for quite some time now :thinking:

bkchr:
I also wanted to have this done as some RFC (I know that I very very slow). The current implementation isn't done in the way it was described in the issue in Substrate.

tomaka:
So correct me if I'm wrong: this PR doesn't actually correspond to #588 (as mentioned in the OP) but corresponds to https://github.com/paritytech/substrate/pull/11490, which has already been deployed ages ago?

bkchr:
> which has already been deployed ages ago

We never activated the host functions, so it was never deployed. So, we can still fix them.

lamafab:
Reverted changes

PR #617, opened 2023-02-19
Fix grammar by switching the and then

HyunggyuJang:
Should be more natural.


FlorianFranzen (approved):
Reads a bit better, thanks.

PR #618, opened 2023-02-19
Fix algorithm Aggregate-Key: $N_i$ to $N$

HyunggyuJang:
$pk_{N_i}$ should be $pk_N$, i.e.,
$$pk^{Agr}_N \leftarrow pk^{Agr}_N || pk_N$$
is the correct formula.


lamafab:
Thank you!

PR #641, opened 2023-05-01
metadata: fix format of constants

FlorianFranzen:
Seems like this was introduced in the "conversion".


0xCaso (approved):
Thanks a lot Florian!

PR #655, opened 2023-06-12
corrections to compact encoding when n>= 2^30

### LINKED ISSUE #654: SCALE: Compact Encoding description needs to be modified
In the Compact/ Length encoding [Definition](https://spec.polkadot.network/id-cryptography-encoding#defn-sc-len-encoding), for the case `n>= 2^ 30`, the value `m` should denote the length of the actual integer being encoded. Currently, m also counts the byte used for encoding length (= the first byte). With this change, the encoding can be used to represent number up to `(2^(63+4)*8) -1 = (2^536) -1`. The implementations already use this convention.

bhargavbh:
resolves https://github.com/w3f/polkadot-spec/issues/654. Definition of compact encoding modified such that m now denotes the lenght of the original integer to be encoded. Added a note to clarify ambiguity. Minor typo fixed for the mode when 2^14 <= n <2^30.

### DIFF
--- a/src/docs/id-cryptography-encoding.md
+++ b/src/docs/id-cryptography-encoding.md
@@ -572,7 +572,7 @@ $$
 {{k}_{{1}}^{{7}}}\ldots{{k}_{{1}}^{{3}}}{{k}_{{1}}^{{2}}}\:={m}-{4}
 $$
 
-Note that ${m}$ denotes the length of the original integer being encoded and does not include the extra-byte describing the length. The encoding can be used for integers upto $$2^{(63+4)8} -1 = 2^{536} -1$$.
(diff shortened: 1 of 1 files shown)


src/docs/id-cryptography-encoding.md
```diff
@@ -572,6 +572,9 @@ $$
 {{k}_{{1}}^{{7}}}\ldots{{k}_{{1}}^{{3}}}{{k}_{{1}}^{{2}}}\:={m}-{4}
 $$
 
+Note that ${m}$ denotes the length of the original integer being encoded and does not include the extra-byte describing the length. The encoding can be used for integers upto $$2^{(63+4)8} -1 = 2^{536} -1$$.
```
0xCaso: Other than this minor typo, lgtm
```suggestion
Note that ${m}$ denotes the length of the original integer being encoded and does not include the extra-byte describing the length. The encoding can be used for integers up to $$2^{(63+4)8} -1 = 2^{536} -1$$.
```

0xCaso (approved):

PR #669, opened 2023-07-11
Issue 650 authority discovery

heversonbr:
This adds the sub-section "Discovering authorities" to the Network Section of the Spec.
It describes the authority discovery mechanism and its interactions with Polkadot nodes and with the DHT layer.


docs/chap-networking.md
```diff
@@ -214,6 +214,158 @@ The Polkadot Host must actively communicate with the network in order to partici
...
+|------------------------|--------------------------------------------------------|
+|  None                  | A type representing `no value`                           | 
+|  HashSet[authorityId]  | An unordered collection of unique `authorityId` elements | 
+
+
+**with**
+- `authorityId` the 256-bit identifier representing the public key of the requested `PeerId`. 
+
+:::
+
```
tomaka: This section describes implementation details of Substrate. This shouldn't be in the spec.
Someone who rewrites a new Polkadot implementation from scratch (Gossamer, Kagome, smoldot, ...) has absolutely no use for any of this.

docs/chap-networking.md
```diff
@@ -214,6 +214,158 @@ The Polkadot Host must actively communicate with the network in order to partici
...
+
+<!-- AUTHORITY DISCOVER interacting with KADEMLIA -->
+#### -sec-num- Publishing local addresses {#sect-auth-discovery-worker-publishing}
+<!-- worker component: network.put_value( hash(authorityId) , SignedAuthorityRecord);  -->
+
+The authority discovery mechanism periodically sends `put_value` messages to the DHT in order to publish a `SignedAuthorityRecord` of the addresses of authorities it knows from its current authority sets. The `put_value` message to be sent to Kademlia DHT is created as follows.
```
tomaka: You can't say that you "send messages to the DHT" in a specification. You can say informally I guess, but it's a vocabulary shortcut.

The Kademlia protocol has an algorithm that describes how to put a value in the DHT, and what Polkadot does is invoke this algorithm. One of the steps in the algorithm is to send messages to specific other nodes. But you don't send messages to the DHT itself, that's non-sensical.

docs/chap-networking.md
```diff
@@ -214,6 +214,158 @@ The Polkadot Host must actively communicate with the network in order to partici
...
+| `bytes`            | 2   | An Schnorrkel/Ristretto x25519 ("sr25519") signature  |
+| *PeerSignature*    | 3   | Serialized peer signature                             |
+
+**where**  
+
+**AuthorityRecord** is a serialized Protobuf structure that lists the addresses of authority nodes that are currently part of the authority set. 
+
+| Type             | Id | Description                                                           |
+|------------------|----|-----------------------------------------------------------------------|
+| `repeated bytes` | 1  | Possibly multiple `MultiAddr`ess through which a node can be connected |
```
tomaka:
```suggestion
| `repeated bytes` | 1  | Binary representation of zero or more multiaddresses through which a node is reachable | 
```

`MultiAddr` is probably how the Substrate struct is named, but the concept is just a "multiaddress", without any special casing.

I would mention specifically the binary representation of a multiaddress, because multiaddresses have both a string and binary representation.

docs/chap-networking.md
```diff
@@ -214,6 +214,158 @@ The Polkadot Host must actively communicate with the network in order to partici
...
+
+**with**
+- `authorityId` the 256-bit identifier representing the public key of the requested `PeerId`. 
+
+:::
+
+
+<!-- AUTHORITY DISCOVER interacting with KADEMLIA -->
+#### -sec-num- Publishing local addresses {#sect-auth-discovery-worker-publishing}
+<!-- worker component: network.put_value( hash(authorityId) , SignedAuthorityRecord);  -->
```
tomaka: You need to mention which hash algorithm and what an `authorityId` means.

docs/chap-networking.md
```diff
@@ -214,6 +214,158 @@ The Polkadot Host must actively communicate with the network in order to partici
...
+
+| Type    | Id  | Description                                       |
+|---------|-----|---------------------------------------------------|
+| `bytes` | 1   | An sr25519 signature                              | 
+| `bytes` | 2   | A sr25519 public key used to verify the signature | 
+
+:::
+
+
+###### Definition -def-num- Publish authority addresses message {#defn-msg-auth-discovery-publish}
```
tomaka: I'm a bit confused by this section. Isn't it the same thing as in the previous section?

docs/chap-networking.md
```diff
@@ -214,6 +214,158 @@ The Polkadot Host must actively communicate with the network in order to partici
...
+
+At each cycle, nodes `worker` creates a bounded number of `get_value` messages to send in the following format:
+$$
+\texttt{get\_value}{\left(\texttt{KademliaKey}_{i}\right)}
+$$
+
+**where**
+- $\texttt{KademliaKey}_{i}$ is the $Sha256$ hash of the `authorityId` of node $i$ selected from the current authority set.
+::: 
+
```
tomaka: Again that's an implementation detail. We don't need to explain how to retrieve the list of authorities, that's an exercise for the reader.

tomaka (commented):
:+1:

docs/chap-networking.md
heversonbr: Although the current implementation relies on substrate, the intention here is actually to describe the Polkadot's AuthorityDiscovery service implementation on Polkadot layer (https://github.com/paritytech/polkadot/blob/4776230c5a03e001fcbea7e930db933cf23a42b7/node/network/protocol/src/authority_discovery.rs#L46) which is, in turn, used by other components of the protocol such as:  bridge's validator discovery, runtimeApiSubsystem, DisputeRequest, gossip-support. However, I agree that mention components that exist on substrate make it confusing, it will be adapted to remove it.

tomaka: The problem is not that it's in Substrate as opposed to Polkadot, the problem is that it's an implementation detail.
There's absolutely no need to conform to anything written in this section in order to be a Polkadot validator, and therefore it shouldn't be in the spec.

docs/chap-networking.md
heversonbr: This has definitely nothing to do with Kad's lookup algorithm and it does not make any reference to it. 
However, I completely understand your point about the term "sending" not being the most suitable description for the periodic publish/lookup (put/get) operations that the discovery mechanism performs on the DHT to ensure up-to-date addresses of authorities. We are taking it into account.

tomaka: > This has definitely nothing to do with Kad's lookup algorithm and it does not make any reference to it.

The algorithm to publish a record is described in the `Value storage` section [here](https://github.com/libp2p/specs/tree/master/kad-dht#value-storage).
Step 1 consists in performing a DHT lookup.

docs/chap-networking.md
heversonbr: Agree.

docs/chap-networking.md
heversonbr: It is mentioned in the Definition

docs/chap-networking.md
heversonbr: This is just the formal definition, part of the Section. I'll rephrase to ensure clarity.

docs/chap-networking.md
heversonbr: I agree that it is indeed straightforward, but providing added convenience for developers could be beneficial.

docs/chap-networking.md
```diff
@@ -214,6 +214,171 @@ The Polkadot Host must actively communicate with the network in order to partici
...
+<!-- POLKADOT NODES interacting with AUTHORITY DISCOVERY -->
+#### -sec-num- Requesting authority identifier and addresses {#sect-auth-discovery-service-requests}
+<!-- `GetAuthorityIdsByPeerId` and `GetAddressesByAuthorityId` are the type of message sent from `service` to `worker` in order to receive the information requested by polkadot nodes. However, this is a implementation detail that can be transparent to developers. 
+ -->
+The following requests are exposed by the discovery authority mechanism to Polkadot nodes.
+
+###### Definition -def-num- Authority addresses request {#defn-auth-discovery-address-request}
+:::definition
+
```
tomaka: (I'm commenting on lines 226 to 306, I have some issues with GitHub's UI)

Again this is not supposed to be in a specification.
The fact that Substrate/Polkadot use this service and worker system is completely an implementation detail.
Could an implementation of Polkadot **not** use this service and worker, or could it use different requests, and still work at 100% of its capacity? Yes, completely. Consequently, not in the specification.

A specification is only to define how implementations of the Polkadot communicate with each other and how they must behave. This service and worker system are how _internal components_ in the source code communicate with each other.

heversonbr: I understand your point, but the text is merely explaining the need for the Authority discovery mechanism to provide two distinguished interfaces and what data is exchanged on each one. The use of names similar to substrate components in the text is purely for illustrative purposes and doesn't imply a direct reference to any specific substrate component. It's simply a choice of terminology to convey the concept.

tomaka: > I understand your point

I don't think you do, and I completely fail to understand your reasoning.

These interfaces are completely arbitrary. Having a service and a worker means that you're using an actor model. Why force client implementations to use an actor model?

It's as if in the specification of the HTTP protocol it was mentioned that the server might have [middlewares](http://expressjs.com/en/guide/using-middleware.html) or uses a database. This would be completely out of place.

heversonbr: While having a service and a worker in a system can be indicative of using an actor model, it is not a definitive indicator on its own. Having a service and a worker could be part of an actor model-based if the components are implemented as independent actors communicating through message passing. However, having a service and a worker could also be part of other architectural patterns or paradigms, such as a client-server, a microservices, or a task-based system, without necessarily adhering to the actor model. The goal here is not to impose any model.

bhargavbh (approved):
LGTM

Noc2 (approved):
Made some small spelling fixes

tomaka:
I know I'm being a pain, but ideally a PR that describes authority discovery shouldn't at the same time make small unrelated changes to half of the networking chapter.
Pull requests are free, there's no need to merge multiple unrelated changes into one PR.

docs/chap-networking.md
```diff
@@ -71,12 +71,12 @@ $$
 
 - ${b}_{{2}}$ and ${b}_{{3}}$ are a protobuf encoded field-value pair [indicating the used key type](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#keys) (field ${1}$ of value ${1}$ implies *ED25519*).
 
-- ${b}_{{4}}$, ${b}_{{5}}$ and ${b}_{{{6}.{.37}}}$ are a protobuf encoded field-value pair where ${b}_{{5}}$ indicates the length of the public key followed by the the raw ED25519 public key itself, which varies for each Polkadot Host and is always 32 bytes (field ${2}$ contains the public key, which has a field value length prefix).
+- ${b}_{{4}}$, ${b}_{{5}}$ and ${b}_{{{6}.{.37}}}$ are a protobuf encoded field-value pair where ${b}_{{5}}$ indicates the length of the public key followed by the raw ED25519 public key itself, which varies for each Polkadot Host and is always 32 bytes (field ${2}$ contains the public key, which has a field value length prefix).
 
 :::
-## -sec-num- Discovery mechanism {#sect-discovery-mechanism}
+## -sec-num- Network bootstrap and discovery {#sect-network-bootstrap}
```
tomaka, line 77: Bootstrapping is nothing more but "discovery but at the very beginning". There IMO should be no need to mention the word "bootstrap" anywhere in the spec, it just adds more vocabulary for no gain.

heversonbr: The choice for " Network bootstrap and discovery" was made to distinguish "Authority discovery" from the "discovery process" to bootstrap the network. Actually, the bootstrap (or bootstrapping process) is a well-defined and accepted definition in the p2p literature and has been used for decades. e,g.,:

- "Bootstrapping of Peer-to-Peer Networks," C. Gauthier Dickey and C. Grothoff, -  International Symposium on Applications and the Internet 2008

- "Bootstrapping large-scale DHT networks." Jae Woo Lee, Henning Schulzrinne, Wolfgang Kellerer, and Zoran Despotovic. 2007. In Proceedings of the 2007 ACM CoNEXT conference 

- "Bootstrapping locality-aware P2P networks," C. Cramer, K. Kutzner and T. Fuhrmann, , Proceedings. 2004 12th IEEE International Conference on Networks 2004

PR #675, opened 2023-07-11
clarify that authority weights are set to 1 in Polkadot

bhargavbh:
Clarifies in definition of Authority List that the weights are set to 1 by the runtime in Polkadot.


Noc2 (changes requested):
I made some additional changes here: https://github.com/w3f/polkadot-spec/pull/676. Feel free to merge both.

PR #682, opened 2023-07-27
BEEFY spec

bhargavbh:
First draft of the BEEFY spec.
TODOs: 
- [x]  describe voting and round selection in more depth
- [x] overall polishing
- [x] Subsampling based light client protocol cna be moved to the Implementer's Guide section as its not part of the core polkadot protocol.


docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
 
 where ${{M}_{{{v},{i}}}^{{\text{Cat}-{s}}}}{\left(\text{id}_{{{\mathbb{{V}}}}},{r}\right)}$ is the catch-up response received from node ${v}$ ([Definition -def-num-ref-](chap-networking#defn-grandpa-catchup-response-msg)).
 :::
+
+
+## -sec-num- Bridge design (BEEFY) {#sect-grandpa-beefy}
+
+:::caution
+The BEEFY protocol is currently in early development and subject to change. The specification is Work-In-Progress.
```
drskalman: So I won't say early development. It s already deployed on Rococo

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+:::
+
+The BEEFY (Bridge Efficiency Enabling Finality Yielder) is a secondary protocol to GRANDPA to support efficient bridging between the Polkadot network (relay chain) and remote, segregated blockchains, such as Ethereum, which were not built with the Polkadot interchain operability in mind. BEEFY’s aim is to efficiently follow a chain that has GRANDPA finality, a finality gadget created for Substrate/Polkadot ecosystem. This is useful for bridges (e.g., Polkadot->Ethereum), where a chain can follow another chain and light clients suitable for low storage devices such as mobile phones.
```
drskalman: BEEFY's aim is to enable client apps to efficiently

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+
+Storing all the information necessary to verify the state of the remote chain, such as the block headers, is too expensive. BEEFY stores the information in a space-efficient way and clients can request additional information over the protocol.
+
+### -sec-num- Motivation {#id-motivation-beefy-1}
+A client could just follow GRANDPA using GRANDPA justifications, sets of signatures from validators. This is used for the substrate-substrate bridge and in light clients such as the Substrate connect browser extension. The main issue with this is that GRANDPA justifications are large and that they are expensive to verify on other chains like Ethereum that do not use the same cryptography. It is not easy to modify GRANDPA to make it better for this. Certain design decisions, like validators voting for different blocks in a justification make this hard. We also don't want to use slower cryptography in GRANDPA. Thus we are adding an extra layer of finality, BEEFY, that will allow lighter bridges and light clients of Polkadot.
```
drskalman: This is used in some substrate-substrate bridge (using *the* both implys that there is only one such instant and it is mandatory to use GRANDPA.

drskalman: It is not easy to modify GRANDPA... I think Spec should have more authoritive tone: GRANDPA is optimiazed for validators to be able to vote for different blocks using fast cryptography and as such it does not allow for such optimization.

And then we should use passive term here: Thus an extra layer of finality, BEEFY has been added to allow lighter bridges and clients for Polkadots.

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+
+To summarise, the goals of BEEFY are:
+- Allow customisation of crypto to adapt for different targets.
```
drskalman: *crypto* is too ambigius here. Perhaps "Signature Algorithm" or "Consensus Votes"? 
*target* -> target chains?

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+
+### -sec-num- Motivation {#id-motivation-beefy-1}
+A client could just follow GRANDPA using GRANDPA justifications, sets of signatures from validators. This is used for the substrate-substrate bridge and in light clients such as the Substrate connect browser extension. The main issue with this is that GRANDPA justifications are large and that they are expensive to verify on other chains like Ethereum that do not use the same cryptography. It is not easy to modify GRANDPA to make it better for this. Certain design decisions, like validators voting for different blocks in a justification make this hard. We also don't want to use slower cryptography in GRANDPA. Thus we are adding an extra layer of finality, BEEFY, that will allow lighter bridges and light clients of Polkadot.
+
+To summarise, the goals of BEEFY are:
+- Allow customisation of crypto to adapt for different targets. 
+- Minimize the size of the "signed payload" and the finality proof.
+- Unify data types and use backward-compatible versioning so that the protocol can be extended (additional payload, different crypto) without breaking existing light clients.
```
drskalman: again different "crypto scheme" or different "signature algorithm".

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+
+### -sec-num- Motivation {#id-motivation-beefy-1}
+A client could just follow GRANDPA using GRANDPA justifications, sets of signatures from validators. This is used for the substrate-substrate bridge and in light clients such as the Substrate connect browser extension. The main issue with this is that GRANDPA justifications are large and that they are expensive to verify on other chains like Ethereum that do not use the same cryptography. It is not easy to modify GRANDPA to make it better for this. Certain design decisions, like validators voting for different blocks in a justification make this hard. We also don't want to use slower cryptography in GRANDPA. Thus we are adding an extra layer of finality, BEEFY, that will allow lighter bridges and light clients of Polkadot.
+
+To summarise, the goals of BEEFY are:
+- Allow customisation of crypto to adapt for different targets. 
+- Minimize the size of the "signed payload" and the finality proof.
```
drskalman: how does BEEFY minimizes the size of "signed payload"  I mean how would the payload be any larger if BEEFY is not used?

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+- Unify data types and use backward-compatible versioning so that the protocol can be extended (additional payload, different crypto) without breaking existing light clients.
+
+### -sec-num- Protocol Overview {#id-overview-beefy-1}
+Since BEEFY runs on top of GRANDPA, similarly to how GRANDPA is lagging behind the best produced (non-finalized) block, BEEFY finalised block lags behind the best GRANDPA (finalised) block. 
+- BEEFY validator set is the same as GRANDPA's, however they might be identified by different session keys.
```
drskalman: I wouldn't use "identified". The session keys on the chain are all attached to the same stash account so that stash account is identifying the validator not the ed25519 GRANDPA key or the BEEFY Key. I'd say "they might be using different type of session key to sign BEEFY messages.

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+
+To summarise, the goals of BEEFY are:
+- Allow customisation of crypto to adapt for different targets. 
+- Minimize the size of the "signed payload" and the finality proof.
+- Unify data types and use backward-compatible versioning so that the protocol can be extended (additional payload, different crypto) without breaking existing light clients.
+
+### -sec-num- Protocol Overview {#id-overview-beefy-1}
+Since BEEFY runs on top of GRANDPA, similarly to how GRANDPA is lagging behind the best produced (non-finalized) block, BEEFY finalised block lags behind the best GRANDPA (finalised) block. 
+- BEEFY validator set is the same as GRANDPA's, however they might be identified by different session keys. 
+- From a single validator perspective, BEEFY has at most one active voting round.
```
drskalman, line 514: Is this accurate? my understanding is that there is no round in BEEFY  (active or not) though I might be wrong. But you don't really care about other people's vote on BEEFY, You see a finalized block (2/3 votes) you are able to sign on it using your BEEFY key, you don't wait for other people,. similarly you see a valid BEEFY vote you Gossip it you don't think which round it is. If don't recollect having round in the BEEFY code.

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+Since BEEFY runs on top of GRANDPA, similarly to how GRANDPA is lagging behind the best produced (non-finalized) block, BEEFY finalised block lags behind the best GRANDPA (finalised) block. 
+- BEEFY validator set is the same as GRANDPA's, however they might be identified by different session keys. 
+- From a single validator perspective, BEEFY has at most one active voting round. 
+- Since GRANDPA validators are reaching finality, we assume they are online and well-connected and have a similar view of the state of the blockchain.
+
+BEEFY consists of two components:
+
+a. **Consensus Extension** on GRANDPA finalisation that is a voting round. 
+
+The consensus extension serves to have smaller consensus justification than GRANDPA and alternative cryptography which helps the second part of BEEFY, the light client protocol described below.
```
drskalman: What do you mean "the second part" of BEEFY? I'd just use the light client side of the beefy protocol it is less confusing.

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+- From a single validator perspective, BEEFY has at most one active voting round. 
+- Since GRANDPA validators are reaching finality, we assume they are online and well-connected and have a similar view of the state of the blockchain.
+
+BEEFY consists of two components:
+
+a. **Consensus Extension** on GRANDPA finalisation that is a voting round. 
+
+The consensus extension serves to have smaller consensus justification than GRANDPA and alternative cryptography which helps the second part of BEEFY, the light client protocol described below. 
+
+b. **Light client** protocol for convincing the other chain/device efficiently about this vote.
```
drskalman: "this vote" → "the finality vote"

docs/sect-finality.md
```diff
@@ -483,3 +483,293 @@ A Catch-up response message contains critical information for the requester node
...
+
+In the BEEFY light client, a prover, a full node or bridge relayer, wants to convince a verifier, a light client that may be on-chain, of the outcome of a BEEFY vote. The prover has access to all voting data from the BEEFY voting round. In the light client part, the prover may generate a proof and send it to the verifier or they may engage in an interactive protocol with several rounds of communication.
```
drskalman: "that may be on-chain" → "which could a bridge implementation on the traget chain"

drskalman (changes requested):
Review of "Bridge design (BEEFY)" section.

docs/sect-finality.md
bhargavbh: here round means choosing which block number to vote on next. You are right that in BEEFY (unlike Grandpa) validators sign when they see 2/3rd votes and do not have to wait. It is just that the the validators have to decide which next block to vote on.

docs/sect-finality.md
bhargavbh: what do you think about "signature schemes"?

docs/sect-finality.md
bhargavbh: what do you think about "signature schemes"?

docs/sect-finality.md
bhargavbh: agreed, i have now changed it "minimise effort required by the light client to follow finality"

docs/sect-finality.md
bhargavbh: rephrased the whole para incorporating this suggestion

docs/sect-finality.md
bhargavbh: good catch, thanks

docs/sect-finality.md
bhargavbh: done

docs/sect-finality.md
bhargavbh: same here, "signature schemes"?

bhargavbh:
@Noc2  Can you please take a look. After your review, can merge this for now. Potentially, the random Subsampling protocol can be added to the implementer's guide in a different PR.

docs/sect-finality.md
```diff
@@ -485,3 +485,268 @@ A Catch-up response message contains critical information for the requester node
...
+
+### -sec-num- Requesting Signed Commitments {#id-requesting-signed-commitments}
+
+A light client ([Definition -def-num-ref-](sect-finality#defn-beefy-light-client)) fetches the Signed Commitment Witness ([Definition -def-num-ref-](sect-finality#defn-beefy-witness-data)) from the chain. Once the light client knows which validators apparently voted for the specified statement, it needs to request the signatures from the relayer to verify whether the claims are actually true. This is achieved by requesting signed commitments ([Definition -def-num-ref-](chap-networking#defn-grandpa-beefy-signed-commitment)).
+
+How those signed commitments are requested by the light client and delivered by the relayer varies among networks or implementations. On Ethereum, for example, the light client can request the signed commitments in the form of a transaction, which results in a response in the form of a transaction.
```
Noc2: Should we keep this: "On Ethereum, for example, the light client can request the signed commitments in the form of a transaction, which results in a response in the form of a transaction."?

Noc2 (changes requested):
Should we keep the Motivation section, or should this be part of the implementer guide?

docs/sect-finality.md
bhargavbh: in my opinion, the 'motivation' section provides context as to why we need BEEFY as an extra layer and why GRANDPA is not optimal for light-clients/bridges. This gives context to the design decisions described later in the chapter. I'm open to changing the title from 'Motivation' to maybe "Requirements", "Objectives"?

We can remove the Ethereum example. Not really needed.

Noc2: Okay. Let's keep it as Motivation

Noc2 (approved):

docs/sect-finality.md
```diff
@@ -485,3 +485,268 @@ A Catch-up response message contains critical information for the requester node
...
+- Parent Block number and Parent Block Hash.
+- Extra Leaf Data: Currently the Merkle root of the list of (ParaID, ParaHeads)
+:::
+
+
+###### Definition -def-num- Commitment {#defn-beefy-Commitment}
+:::definition
+
+**Signed Commitment:**
+`commitment` is a tuple of `(payload, Block Number, ValidatorSetID)`. A `signed commitment` is a tuple `(commitment, signatures)`, where `signatures` is a list of optional signatures of the validator set on the SCALE encoded `commitment`. Note that the number of signatures in `signatures` may be less than the length of the Validator Set.
```
heversonbr, line 619: The Definition is called "Commitment" but what is in fact defined is a "Singed Commitment".  Wouldn't it be better to either divide it into two distinguished definitions, or simply call that "Signed Commitment" and make the definition of commitment an inner preliminary definition of

heversonbr: The Definition is called "Commitment" but what is in fact defined is a "Singed Commitment". Wouldn't it be better to either divide it into two distinguished definitions, or simply call that "Signed Commitment" and make the definition of commitment an inner preliminary definition of it ?

docs/sect-finality.md
```diff
@@ -485,3 +485,268 @@ A Catch-up response message contains critical information for the requester node
...
+:::
+###### Definition -def-num- Relayer {#defn-beefy-relayer}
+:::definition
+
+A **relayer** (or "prover") is an abstract entity that takes finality proofs from the Polkadot network and makes those available to the light clients. The relayer attempts to convince the light clients that the finality proofs have been voted for by the Polkadot relay chain validators. The relayer operates off-chain and can for example be a node or a collection of nodes.
+
+:::
+### -sec-num- Voting on Statements {#id-voting-on-statements}
+
+The Polkadot Host signs a statement ([Definition -def-num-ref-](sect-finality#defn-beefy-statement)) and gossips it as part of a vote ([Definition -def-num-ref-](chap-networking#defn-msg-beefy-gossip)) to its peers on every new finalized block. The Polkadot Host uses ECDSA for signing the statement since Ethereum has better compatibility for it compared to SR25519 or ED25519.
```
heversonbr: The statement definition seems to be missing. Also, the Beefy gossip message does not seem to be defined.

docs/sect-finality.md
```diff
@@ -485,3 +485,268 @@ A Catch-up response message contains critical information for the requester node
...
+
+### -sec-num- Committing Witnesses {#sect-beefy-committing-witnesses}
+
+The relayer ([Definition -def-num-ref-](sect-finality#defn-beefy-relayer)) participates in the Polkadot network by collecting the gossiped votes ([Definition -def-num-ref-](chap-networking#defn-msg-beefy-gossip)). Those votes are converted into the witness data structure ([Definition -def-num-ref-](sect-finality#defn-beefy-witness-data)). The relayer saves the data on the chain of the remote network. The occurrence of saving witnesses on remote networks is undefined.
+
+### -sec-num- Requesting Signed Commitments {#id-requesting-signed-commitments}
+
+A light client ([Definition -def-num-ref-](sect-finality#defn-beefy-light-client)) fetches the Signed Commitment Witness ([Definition -def-num-ref-](sect-finality#defn-beefy-witness-data)) from the chain. Once the light client knows which validators apparently voted for the specified statement, it needs to request the signatures from the relayer to verify whether the claims are actually true. This is achieved by requesting signed commitments ([Definition -def-num-ref-](chap-networking#defn-grandpa-beefy-signed-commitment)).
```
heversonbr: Same here for : chap-networking#defn-grandpa-beefy-signed-commitment

docs/sect-finality.md
```diff
@@ -485,3 +485,268 @@ A Catch-up response message contains critical information for the requester node
...
+session.
+
+While it is useful to finalize non-mandatory blocks frequently, in good networking conditions BEEFY may end up finalizing each and every block GRANDPA finalized block. Practically, with short block times, it's going to be rare and might be excessive, so
+it's suggested for implementations to introduce a `min_delta` parameter which will limit the frequency with which new rounds are started. The affected component of the formula would be:
+`best_beefy + MAX(min_delta, NEXT_POWER_OF_TWO(...))`, so we start a new round only if the
+the power-of-two component is greater than the min delta. Note that if `round_number > best_grandpa` the validators are not expected to start any round.
+
+### -sec-num- BEEFY Light Client {#id-light-client-beefy-1}
+
+A light client following BEEFY could request $N/3 +1$ signatures to be checked, where $N$ is the number of validators on Polkadot. Assuming a maximum of $N/3$ malicious validators, the light client can be certain of the payloads finality if all the signatures it requested are valid.
```
heversonbr, line 765: this sub-section shouldn't be somehow in the Light clients chapter ? 
I may be wrong but it seems to be more an explanation of requirements than a specification itself.

heversonbr (changes requested):

docs/sect-finality.md
bhargavbh: i have changed the definition header, thanks

docs/sect-finality.md
bhargavbh: Yes, it is just a place holder. Ideally we move some of this with an instantiation of a light client (eg. the Ethereum<>Polkadot bridge) to the implementer's guide.

docs/sect-finality.md
bhargavbh: i have added the missing links and definitions

docs/sect-finality.md
bhargavbh: added the links and definitions in the latest push

Noc2 (approved):

heversonbr (approved):

bhargavbh:
I suggest merging this PR for now. How the light client verifies (via interactive Random Subsampling of signatures) BEEFY payload can be part of the implementers guide in a separate PR.

Issue #681, opened 2023-07-24
BABE skipped epochs


andresilva:
Whenever there is a period of inactivity (i.e. no blocks being produced) for more than one epoch BABE is broken, since we announce epochs in advance and by the time the inactivity is resolved we will be in a future epoch which was unnanounced (i.e. no authorities and randomness available). The security model of BABE is also broken in this scenario, but for a long time in the Substrate implementation at least, this would completely brick the chain and make it impossible to author new blocks. As of https://github.com/paritytech/substrate/pull/11727 this was changed, such that empty epochs can be skipped over, i.e. after a period of inactivity the runtime will appropriately update the epoch index based on the given slot. We are still left with the problem of what epoch data to use and the natural thing to do is to re-use the data for the latest announced epoch. As an example, assuming the last block was authored on a slot at epoch 3, the data for epoch 4 had already been announced. Given a period of inactivity spanning 3 epochs, a block is then authored with a slot belonging to epoch 6. In order to author this block, the client should re-use the data that was previously published for epoch 4.

andresilva:
This problem is also described by the Ouroboros team in [this](https://api.zotero.org/groups/478201/items/RR93NMSD/file/view?key=PnpP8O1NApZMMF0LVNh7I4I5) paper in chapter 2. They describe a manual mitigation which is the same as we have employed in early 2020 on [Kusama](https://polkadot.network/blog/kusamas-first-adventure) (we never had to do it on Polkadot mainnet). They do not provide any description of an automatic mitigation method though.

Noc2:
Thanks a lot for creating the issue! We will look into it as soon as possible.

hndnklnc:
Do we need a solution for only one empty epoch or more than one sequential epochs?

tomaka:
Does the first header after a period of inactivity contain an epoch change digest? My understanding is yes.

For example, let's say that we're in the middle of epoch 3, and all of a sudden all the validators die. Two weeks later, they come back online.
Is it correct that the first authored block is in epoch 9 (or whatever the number is) and announces epoch 10 (or whatever the number is)? And epoch 9 uses the authorities and randomness of epoch 4?

andresilva:
@tomaka Everything you said is correct.

[...]

PR #685, opened 2023-08-08
BABE: added reusing last epoch data in case of skipped epochs

bhargavbh:
Addresses https://github.com/w3f/polkadot-spec/issues/681. 
In case BABE fails to produce blocks for epochs $E_{n+1}$ to $E_{n+k}$, the epoch data announced in $E_n$ is used for producing blocks at $E_{n+k+1}$. This detail has been added in the Block Production [section](https://spec.polkadot.network/sect-block-production#defn-consensus-message-babe)


andresilva (approved):
Definition in spec looks good to me. Although the definition in the description of the PR is a bit ambiguous to me, what is meant by `epoch data En`? The epoch data that was announced at the beginning of `En` and that was meant to be used in `En + 1`? If that's the case then I agree.

Also unrelated to the changes here but in section 5.3 in the `Danger` box, slot calculation is defined as `s = Tunix * slot_duration` whereas it should be `s = Tunix / slot_duration`.

bhargavbh:
thanks, i updated the PR description and fixed the typo.

PR #691, opened 2023-08-28
Fix typo "Polkodat" in AnV chapter

laboon:


Noc2 (approved):
Thanks for the PR.... we were secretly working on the Polkodat Network, but I guess now it's no longer a secrete ;-)

Issue #692, opened 2023-08-28
Controller accounts are being deprecated


laboon:
I noticed several references and details to Controller accounts (e.g. https://spec.polkadot.network/id-cryptography-encoding ). Note that these are in the process of being deprecated ( https://forum.polkadot.network/t/staking-controller-deprecation-plan-staking-ui-leads-comms/2748 ) and these will need to be removed.

heversonbr:
Fixed on Controller keys deprecation [#700](https://github.com/w3f/polkadot-spec/pull/700)

PR #700, opened 2023-09-07
Controller keys deprecation

heversonbr:
Tagged definition as deprecated and  removed controller keys information from other parts of the text due to deprecation

Handles issue:  https://github.com/w3f/polkadot-spec/issues/692


bhargavbh (changes requested):
maybe link to the wiki explaining the stashing keys can be added at the appropriate place?

Noc2 (approved):
I removed: "The Stash key holds the majority of the users’ funds and should neither be shared with anyone, saved on an online device, nor used to submit extrinsics." (doesn't need to be part of the spec). Otherwise, feel free to merge

Noc2 (approved):

bhargavbh (approved):

PR #703, opened 2023-09-12
fix genesis state

tjjfvi:
At least for Polkadot, the genesis header's `extrinsic_root` is not set to `[0; 32]`, but rather `blake2b([0])` (see e.g. [here](https://explorer.polkascan.io/polkadot/block/0)).

This PR also clarifies the size of the zero values for the other fields (with notation used elsewhere in the spec) – this is subjective, but I feel it is clearer this way. That change is made in a separate commit, though, so it can be easily removed.


Noc2 (approved):
Thanks a lot for the update and contribution ([confirmed](https://polkadot.subscan.io/block/0))

bhargavbh (approved):
LGTM, thanks

PR #705, opened 2023-09-26
fix: missing information about SCALE encoding on HttpRequestStatus codes

heversonbr:


bhargavbh (approved):

docs/chap-host-api.md
```diff
@@ -1088,7 +1088,7 @@ All data and results generated by Offchain workers are unique per node and nonde
 ###### Definition -def-num- HTTP Status Code {#defn-http-status-code}
 :::definition
 
-**HTTP status codes** that can get returned by certain Offchain HTTP functions.
+A SCALE encoded structure holding of one of the following **HTTP status codes** that get returned by offchain http functions.
```
drskalman: 1. Do we define 'struct' type and how SCALE encodes it?. I remember it was called tuple or something.
2. this is rather varying data type because only in case of Finish you have an extra field.
3. Maybe we should refer to the SCALE section which ever the type is.
4. Does by `PassByCodec` actully convert `HttpRequestStatus` to `u32` calling `from` defined here: https://github.com/paritytech/polkadot-sdk/blob/cfc6cc65cc133f69f472149b2449a673597e8530/substrate/primitives/core/src/offchain/mod.rs#L160 ? i suspect it doesn't and it just encodes: 0,1,2,3 ? that might be why @tomaka  says "In fact 10 and 20 don't even correspond to anything" Maybe @tomaka could clarify.

bkchr: `PassByCodec` means that we call `encode()` on the type before putting it into a `u64` (ptr + len). As the type is a Rust enum, it follows the default SCALE enum encoding.

heversonbr: > 1. Do we define 'struct' type and how SCALE encodes it?. I remember it was called tuple or something.
    You are right. 
> 2. this is rather varying data type because only in case of Finish you have an extra field.
    You are right.
> 3. Maybe we should refer to the SCALE section which ever the type is.
    I agree. The current definition of a varying data type matches the 'default SCALE enum encoding' mentioned by @bkchr.  I'll change the definition using the already defined operation.

bkchr (approved):
The representation is now correct! Ty!

bhargavbh (approved):

Noc2 (approved):

PR #715, opened 2024-01-16
Updated "Fast Sync" section

Imod7:
### Description
As it is stated right now in the "Fast Sync" section, it is unclear, once fast sync is completed, if : 
- the node will automatically switch to full sync 
- or the node operator needs to stop the node and start it again with the flag `--sync=full`

So, we think it is better to explicitly mention this fact in order to avoid any confusion.

Credits to @altonen for the feedback!

cc: @IkerAlus


altonen (approved):

IkerAlus (approved):

Noc2 (approved):

bhargavbh (approved):
thanks!

PR #734, opened 2025-01-13
Fix Best Pre-Vote Candidate in GRANDPA Spec

### LINKED ISSUE #733: GRANDPA pre-vote selection algorithm
The spec for prevoting is inconsistent with Substrate and the GRANDPA paper. 

Best-PreVote-Candidate returns either the primaries block or the ghost of the prevotes apparently from this round r.

What the validator should prevote for is for the last audited block (or 3 blocks from the end in single-chain Substrate) in the longest chain (or chain with the most primary blocks for BABE) including a block B defined as follows

If the primaries' block is between the last finalused block L and the ghost of the prevotes from round r-1, then that is B 

else use the ghost of the prevotes from round r-1.

We should probably check whether Substrate works like this.

drskalman:
closes #733

- Best Prevote should come from last round's GHOST.
- We should return the best chain safe head of that GHOST.

### DIFF
--- a/algorithms/bestPrevoteCandidate.tex
+++ b/algorithms/bestPrevoteCandidate.tex
@@ -2,11 +2,12 @@
 \caption{Best-PreVote-Candidate}
 \begin{algorithmic}
     \input $r$
-    \state $B^{r, pv}_v \leftarrow$ \call{GRANDPA-GHOST}{$r$}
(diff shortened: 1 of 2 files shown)


bhargavbh (approved):
LGTM. Do we need to make any changes in the BABE section s.t it is consistent with the "Best Safe Chain Head" terminology?

Noc2 (approved):

